PR #43, opened 2012-09-19
Added XRegExp.join() which joins an Array with the given separator in… 

KylePDavis:
...a single XRegExp

Split the majority of XRegExp.union into a private prepareJoin() helper.
Added XRegExp.join() which joins an Array with the given separator into
a single XRegExp.
Updated XRegExp.union() to use new XRegExp.join() internally.
Rebuilt the xregexp-all.js file.


slevithan:
Closing in favor of #106.

Issue #78, opened 2014-10-10
Could you please update xregexp on npm and bower to 3.0


andyhu:
Thanks!

EgoAleSum:
+1
The workaround works but every time you run "npm update" it re-downloads everything. 3.0 seems pretty stable to me.

riquito:
In the meanwhile do not use the master's tarball, use a fixed commit hash, it will download the data only once.

e.g. in bower

    "dependencies": {
        "xregexp": "slevithan/xregexp#2650989f9a"
    }

slevithan:
Thanks for all the interest and support. There are some as-yet-incomplete backward incompatible changes I'm planning to make for 3.0, which is why I've been reluctant to update it until getting back into the project. The release of Unicode 8.0 final soon might be a good time for this.

gajus:
[Unicode 8.0.0](https://www.npmjs.com/package/unicode-8.0.0) has been released now.

slevithan:
Yup! A number of improvements have been made since last commenting here, and XRegExp v3.0.0 is finally nearing release. Want to make a couple final tweaks to support ES6 regex changes and update the Unicode data, and it will ready.

slevithan:
XRegExp 3.0.0 final was released on GitHub and npm a week ago now. Will register on Bower shortly.

Issue #63, opened 2014-02-01
3.0 on npm


whitecolor:
It seems that npm now has only 2.0., are you going to publish 3.0?

slevithan:
Will use #91 to cover adding bower.json.

PR #91, opened 2015-07-28
bower.json

tomascassidy:
Add bower.json to support bower.io

Most fields were copied from package.json but wasn't sure what to add for ```dependencies``` and ```devDependencies```.  Explanation for what to put in those fields is at http://bower.io/docs/creating-packages/

You will also need to register the XRegExp package with bower once bower.json has been added, using the instructions provided at the above link.

#63 #78

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,34 @@
+{
+    "name": "xregexp",
+    "main": "xregexp-all.js",
+    "authors": [
+        "Steven Levithan <steves_list@hotmail.com>"
+    ],
+    "description": "Extended regular expressions",
+    "dependencies": {},
+    "devDependencies": {},
+    "moduleType": [
+        "amd",
+        "globals",
+        "node"
+    ],
+    "homepage": "http://xregexp.com/",
+    "license": "MIT",
+    "keywords": [
+        "regex",
+        "regexp"
+    ],
+    "ignore": [
+        "**/.*",
+        "package.json",
+        "README.md",
+        "src",
+        "tests",
+        "tools"
+    ],
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/slevithan/xregexp.git"
+    }
+}
+


listepo:
+1

slevithan:
Planning to add this shortly and register with bower, hopefully this weekend. First want to read the above referenced doc and play with Bower a bit. Thanks, @tomascassidy!

slevithan:
I've registered `xregexp` with Bower. @tomascassidy and @listepo do things seem to be working as expected? I have not used Bower previously.

tomascassidy:
I tried installing ```xregexp``` with bower in a new empty directory, with default options for ```bower init```, but it appears to be ignoring the ```ignore``` option in bower.json.  I'm still seeing all the files in the repo get installed despite them appearing in the ```ignore``` section.  Can you verify this?  I'm not an export on bower so not really sure what's going on here.  Don't worry too much if it can't be fixed up.

slevithan:
That makes sense actually. Since I didn't know how Bower worked, I screwed up by not merging your `bower.json` until after publishing the `v3.0.0` git tag. :-( And because the latest tag does not contain a `bower.json`, it's not ignoring the `ignore` section; rather there's no `bower.json` to follow despite the project now being registered with Bower. This will be fixed as soon as I publish a new semver git tag.

In order to address this sooner, I'll publish a `v3.0.1` tag as soon as I fix a bug or two.

tomascassidy:
Thanks.  That looks like it should fix what I was seeing.

Goddak:
@slevithan do you have an ETA on when you're able to publish a new semver? I'm relying on a couple of properties added in the `bower.json` in d0e665d4068cec4d15919215b098b2373f1f12e9 for my build tooling.

slevithan:
@Goddak I've been slower on this than I expected, but I'm planning to publish an update within the next couple weeks.

Goddak:
+1 @slevithan

slevithan:
v3.1.0 was released today. :dancer:

Issue #96, opened 2015-08-26
Improve XRegExp.matchRecursive handling of unbalanced delimiters


farfromrefug:
Hi,

First thanks for your great library.
I am using matchRecursive to find JSON messages on a stream.
Every time i receive a message, i push it a string stack.
Then i use `matchRecursive` to try and find messages. If i find some, i slice those messages from the stack.
Now sometimes some messages get lost and i end up with "unbalanced" stack, like this:
`{"toto":"test","bd":{"toto":"test","bd":{"toto":"test","bd":{"toto":"test","bd":1,"asdfg":true}{"toto":"test","bd":{"toto":"test","bd":1,"asdfg":true}{"toto":"test","bd":1,"asdfg":{"toto":2}}`

If i use your library like so i cant get the message in this because it's unbalanced.
And this is correct. there is no `outputs` if you look at this message.

Though there is a simple way to find the messages. Reverse the string and inverse the left and right delimiters. 
It works and `matchRecursive` actually found the message. However it still reports unbalanced and throw at the end.
I easily fixed this locally by only throwing if `ouptuts.length===0` like this:

```
if (output.length === 0) {
                    throw new Error("string contains unbalanced delimiters");
                } else {
                    return output;
                }
```

That way i can still get the matched strings even if at the end it is unbalanced.
This is very useful.

I was wondering if you would be willing to commit that in the master branch?
I think it would be a great addition (could become an option).

slevithan:
Yes, I'd be totally willing to include an option that allows `XRegExp.matchRecursive` to skip past unbalanced delimiters. In fact, I'd prefer to make that the default. Perhaps the new option could be called `requireBalance`, and when set to `true` it would revert to the current handling where scanning past an unbalanced delimiter throws an error.

Unfortunately, I don't expect to get to this for XRegExp 3.0 final, unless it's provided as a pull request.

The challenge is supporting all of `matchRecursive`'s existing features and providing robust tests proving the change is safe. Right now `matchRecursive` is under-tested. See https://github.com/slevithan/xregexp/blob/master/tests/spec/s-addons-matchrecursive.js which currently just tests the examples from the readme, rather than testing individual features, combinations, and assumptions.

Features that would need to continue working correctly after such a change is introduced include:

* Sticky and non-sticky mode, via flag `y`.
* Global and non-global mode, via flag `g`.
* The `escapeChar` option.
[...]

PR #99, opened 2015-08-28
return existing results if unbalanced

farfromrefug:


slevithan:
See responses on #96. I think this might be too narrow in its handling.

slevithan:
Unfortunately this can't be supported in its current form.

Issue #45, opened 2012-11-28
Using named group object is difficult


Ciantic:
I need to loop over the named parameters after executing, seems to be pretty problematic. Why are you augmenting the array with properties? Shouldn't you provide a way to give access to clean named object?

Taken regex `"(?<type>[a-z]+)/(?<id>\\d+)/(?<tab>[a-z]*)?"` I'd like to have object (after exec):

```
{ 'type' : 'bar', 'id' : 666, 'tab' : 'ifany' }
```

That would be trivial to loop over, now I just can't loop over the properties normally cause there is crap preceding the properties (and even worse, if there is update to XRegExp I might get new crap to skip over in loop).

vsimonian:
This issue is a huge show-stopper before I'm able to utilize XRegExp (I don't mean to sound pompous, but it really is a big problem in my use-case).

In the following example, `matches` is, as expected, an array, with built-in array functions, and all the capture data.

```javascript
var regex = XRegExp('(?<cap>[a-z]+)')
var matches = XRegExp.exec('data', regex)
matches.cap === 'data'
```

This becomes problematic when named capture groups coincide with any standard array properties, as the results will override the array property/member.

In this example, the native `slice` function is replaced with the contents of the `slice` named capture group.

```javascript
var regex = XRegExp('(?<slice>[a-z]+)')
var matches = XRegExp.exec('data', regex)
matches.slice === 'data'
```

It simply isn't manageable to have to tiptoe around these names, especially when developing software which exposes XRegExp's functionality outwards in one fashion or another. It's unreasonable to have the expectation that every user will avoid every array property name in their regular expressions. Such arbitrary restrictions reveal a design flaw.

At the very least, there needs to be namespacing, i.e.:

```javascript
var regex = XRegExp('(?<slice>[a-z]+)')
var matches = XRegExp.exec('data', regex)
matches.named.slice === 'data'
```

Also, whatever happened to #51? Is `.names` or an equivalent still planned? It seems an essential feature, yet it isn't implemented as of yet in 3.0.0-pre though the issue is over several years old (and this one is older).

vsimonian:
Now that 3.0.0 is released, is there any plan for namespacing capture groups? It would be a breaking change, and 3.1.0, as a minor version update, cannot include breaking changes.

Perhaps, to mitigate against breaking default behaviour, there could be an option that enforces namespaced results.

slevithan:
> Is `.names` or an equivalent still planned?

[...]

PR #104, opened 2015-09-22
Implement .names

vsimonian:
Per https://github.com/slevithan/xregexp/issues/45#issuecomment-142207025

> - Add `<regexp>.xregexp.names` to all XRegExps. `names` is an array of 0+ strings holding the names of named capturing groups used by the regex, unlike `captureNames`' interspersed `null`s for unnamed capturing groups and `null` value if there are no named groups.
> - Rename `captureNames` as `_captures`. This is technically a non-breaking change because `<regexp>.xregexp.captureNames` is undocumented.
> - Ensure a copy of the `names` array is added to regexes copied via `XRegExp(regexp)` and `XRegExp.globalize(regexp)`.
> - Ensure that `<regexp>.xregexp.names` is available on the regex passed as the 4th argument to `XRegExp.forEach` callbacks.
> - Add tests for all of the above.

### DIFF
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -40,7 +40,7 @@
  */
     function asXRegExp(value) {
         return XRegExp.isRegExp(value) ?
-            (value[REGEX_DATA] && value[REGEX_DATA].captureNames ?
+            (value[REGEX_DATA] && value[REGEX_DATA]._captures ?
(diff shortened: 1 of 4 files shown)


slevithan:
Closing in favor of the approach described in https://github.com/slevithan/xregexp/issues/45#issuecomment-292875730

PR #113, opened 2015-10-10
fix AMD detection

chocolateboy:
The Firefox Addon SDK defines a global `define` function, which XRegExp's UMD stanza mistakes for AMD. Adding a check for `define.amd` fixes this and is the standard way to detect AMD:

https://github.com/umdjs/umd/blob/3e14a7dfb0af2bd3a8bfe9b7b384d698e3ef75e4/amdWeb.js#L19

### DIFF
--- a/tools/intro.js
+++ b/tools/intro.js
@@ -8,7 +8,7 @@
 // Don't use strict mode for this function, so it can assign to global
 ;(function(root, definition) {
     // RequireJS
-    if (typeof define === 'function') {
+    if (typeof define === 'function' && define.amd) {
         define(definition);
     // CommonJS
     } else if (typeof exports === 'object') {
--- a/xregexp-all.js
+++ b/xregexp-all.js
@@ -8,7 +8,7 @@
 // Don't use strict mode for this function, so it can assign to global
 ;(function(root, definition) {
     // RequireJS
-    if (typeof define === 'function') {
+    if (typeof define === 'function' && define.amd) {
         define(definition);
     // CommonJS
     } else if (typeof exports === 'object') {


mathiasbynens:
cc @jdalton

d10:
Lodash currently uses the following in order to allow optimization by AMD build optimizers:

```js
  // Some AMD build optimizers like r.js check for condition patterns like the following:
  if (typeof define == 'function' && typeof define.amd == 'object' && define.amd) {
```

However, the current simplified version in this PR is also recognised by [r.js](https://github.com/jrburke/r.js/blob/master/build/jslib/pragma.js#L39), so I guess it’s a matter of preference on which version to use.

chocolateboy:
@mathiasbynens AFAICT, lodash v4 (under construction) works fine with my Firefox addon (under construction):

https://github.com/lodash/lodash/issues/223#issuecomment-142839054

@d10 Should I add that extra clause to this PR (and push it back to the UMD repo)? Is there an environment in which we need to add:

```
&& define.amd == 'object'
```

to avoid a false positive?

d10:
AFAIK there isn’t a need to do that, but @jdalton will probably know better ;)

chocolateboy:
cc @addyosmani

jrburke:
define.amd is explicitly in the AMD API to help avoid false positives. If the Firefox Addon SDK exposes a define() but it does not operate like an AMD define, then this is a great reason to add the detection. The patterns at https://github.com/umdjs/umd all use it to avoid the false positives, I suggest using it in the detection.

jdalton:
> && typeof define.amd === 'object'

It was just preference for being explicit. Use whichever you prefer and is supported ;)

slevithan:
Thanks for the report and the pull request, @chocolateboy!

Can you update the pull request to not modify `tests/perf/versions/xregexp-all-v3.0.0.js`? That's the already tagged/released v3.0.0. I'll include this fix in a v3.0.1 or v3.1.0.

chocolateboy:
> Can you update the pull request to not `modify tests/perf/versions/xregexp-all-v3.0.0.js`?

Done.

Issue #112, opened 2015-10-05
`(?:)*` should not be a syntax error


slevithan:
Reported by Jan Goyvaerts over email. As of XRegExp 3.0.0, a leading `(?:)*` is a syntax error. This is due to the cleanup at [xregexp.js#L614-L615](https://github.com/slevithan/xregexp/blob/7563982966f8606bfe79c3698efd4e355e2d775e/src/xregexp.js#L614-L615).

`(?:)*` serves no purpose, so this is a low-priority bug. However, it was an unintended change in XRegExp 3.0.0.

Note that `XRegExp.build` has special handling for leading `^` and trailing `$` that relies on XRegExp's cleanup of leading and trailing empty `(?:)` groups, since various XRegExp syntax tokens and flags inject `(?:)` to ensure that preceding and following tokens are kept independent from each other.

A possible fix for this is to strip any immediately following quantifier along with removed empty `(?:)` groups. Alternatively, XRegExp could be more careful about where `(?:)` is injected, or not rely on cleanup of empty noncapturing groups.

jfahrenkrug:
I've created a fix for it here: https://github.com/slevithan/xregexp/pull/116

PR #116, opened 2015-10-22
GH-112 Allow (?:)*/+/?/{1} without causing a syntax error

jfahrenkrug:
This fixes https://github.com/slevithan/xregexp/issues/112

The new cleanup regex also cleans up `*`, `+`, `?` or `{\d}` after an empty capture group. 
You can see a demo here: http://rubular.com/r/AFFiDroJEg

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -612,7 +612,7 @@ var XRegExp = (function(undefined) {
 
             patternCache[pattern][flags] = {
                 // Cleanup token cruft: repeated `(?:)(?:)` and leading/trailing `(?:)`
-                pattern: nativ.replace.call(output, /\(\?:\)(?=\(\?:\))|^\(\?:\)|\(\?:\)$/g, ''),
+                pattern: nativ.replace.call(output, /\(\?:\)(?:[*+?]|\{\d+(?:,\d*)?})?\??(?=\(\?:\))|^\(\?:\)(?:[*+?]|\{\d+(?:,\d*)?})?\??|\(\?:\)(?:[*+?]|\{\d+(?:,\d*)?})?\??$/g, ''),
                 // Strip all but native flags
                 flags: nativ.replace.call(appliedFlags, /[^gimuy]+/g, ''),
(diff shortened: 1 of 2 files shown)


slevithan:
Thanks, @jfahrenkrug! That's a nice improvement.

I've noticed in testing that it still gets some cases wrong, and also escapes a number of characters that don't need to be escaped. I've updated your regex and test cases at <http://rubular.com/r/0G9Dr76yb6>. Note that even the updated regex still gets the last example wrong though (it matches `(?:)` at the end of the string even if the `(` is escaped), but then the current cleanup regex also has this problem.

If you update the PR to use the regex at <http://rubular.com/r/0G9Dr76yb6> and add a few more test cases, I'd be happy to merge it.

(The case that this still gets wrong points to a better fix being needed in the longer term, though.)

jfahrenkrug:
Thanks, @slevithan!

I've updated the regex but I haven't gotten around to the tests yet. Could you give me one example of the tests you'd like to see?

Thanks!

slevithan:
@jfahrenkrug, looking again, I think the existing tests you added are good. Maybe just add the remaining types of quantifiers: `{1,}`, `{1,2}`, `??`, `*?`, etc. I'd also recommend changing `.not.toThrowError(SyntaxError)` to `.not.toThrow()`, since we don't want it to throw any kind of error.

Would you mind squashing your updates into a single diff? https://help.github.com/articles/about-git-rebase/ has details on how to do that.

jfahrenkrug:
@slevithan Thank you! I'll take care of it and make the changes.

jfahrenkrug:
@slevithan The commits are squashed and should be good to go :)

slevithan:
Thanks, @jfahrenkrug!

slevithan:
Note that in commit 69e552b5ecba6769bbc97bd369115c487439a69d I've changed the strategy here by removing the cleanup regex and instead just dealing directly with where this can cause a problem in `XRegExp.build`. This also fixes the longstanding issue of stripping an escaped trailing `\(?:)`.

PR #118, opened 2015-11-04
allow \x as alternative to \u for unicode code points 

bcoe:
As outlined here:

http://www.regular-expressions.info/unicode.html

`\x` is used as an alternative to `\u' in some regex engines (I'm specifically thinking of [oniguruma](https://github.com/rubinius/oniguruma)) which I'm working on writing some shims for.

It would be nice if `\x{xxxx}` was accepted as an alternative to `\u{xxxx}` I can work around this with an add on for the time being :+1:

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -1701,7 +1701,7 @@ var XRegExp = (function(undefined) {
  * if you use the same in a character class.
  */
     add(
-        /\\u{([\dA-Fa-f]+)}/,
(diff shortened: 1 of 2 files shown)


slevithan:
XRegExp's philosophy is generally not to add redundant syntax unless it provides some additional functionality. However, `\x{...}` is common enough (it's also used by Perl and PCRE) that you're probably right about it being worth adding in XRegExp. However, I don't expect to add this in a minor version, since syntax changes requires tools that parse XRegExp syntax (e.g. RegexBuddy) to update.

PR #122, opened 2015-12-11
Updates the package to use CommonJS and Browserify for the build

alexgorbatchev:
This makes the `XRegExp` package a proper citizen on the NPM infrastructure while maintaining `xregexp-all.js` compatibility.


alexgorbatchev:
@slevithan could you have a look at this PR please?

slevithan:
@alexgorbatchev I support this in principle but want to take some time to look at it more closely and test. Might have some minor change requests as well.

Can you explain a little more about the changes and why they're needed or helpful.

alexgorbatchev:
The changes are focused around stripping out the code that only used to creating local scope necessary during concatenation process. The concatenation is replaced with CommonJS `require()` calls which creates own scope for each module.

I added `index.js` that serves as the entry point for the module which then reexports `XRegExp` and applies all addons to it at the same time.

I then removed everything that was used to create concatenated file and replaced it with a single `browserify` command that takes the entry `index.js` and resolves all of the `require()` statements creating a single bundle. I added a `--standalone` to mimic how `XRegExp` was exported on the `window` originally.

The last part was just to update the test file to point to the `xregexp-all.js` instead of each file individually, no changes were necessary anywhere else and every test is passing as expected.

This is a modern way of publishing NPM modules and building modular javascript libraries and applications.

My personal interest in this is that I want to be able to link to `xregexp.js` without all the addons from my library to save on the filesize. If this is merged, I'll be able to do a simple `require('xregexp/src/xregexp')` and just get what I need.

slevithan:
@alexgorbatchev Thanks for the details.

@mathiasbynens Any thoughts on this before I merge it?

src/addons/build.js
```diff
@@ -169,4 +169,4 @@
         return XRegExp(pattern, flags);
     };
 
-}(XRegExp));
+}
```
mathiasbynens: missing `;`

src/addons/matchrecursive.js
```diff
@@ -188,4 +188,4 @@
         return output;
     };
 
-}(XRegExp));
+}
```
mathiasbynens: missing `;`

src/addons/unicode-base.js
```diff
@@ -233,4 +233,4 @@
         XRegExp.cache.flush('patterns');
     };
 
-}(XRegExp));
+}
```
mathiasbynens: missing `;`

src/addons/unicode-blocks.js
```diff
@@ -1071,4 +1071,4 @@
         }
     ]);
 
-}(XRegExp));
+}
```
mathiasbynens: missing `;`

src/addons/unicode-scripts.js
```diff
@@ -556,4 +556,4 @@
         }
     ]);
 
-}(XRegExp));
+}
```
mathiasbynens: missing `;`

alexgorbatchev:
Fixed missing `;`

alexgorbatchev:
Awesome, thank you fellas!

Issue #121, opened 2015-12-07
Use of RegExp.prototype getters to determine support is invalid


marcins:
In Chrome Canary 49.0.2583.0 with "Experimental JavaScript" enabled XRegExp throws an error because it tries to read `RegExp.prototype.sticky`.  Canary's behaviour is actually correct according to the ES2015 spec: http://www.ecma-international.org/ecma-262/6.0/#sec-get-regexp.prototype.sticky

An alternative method should be use to determine whether native support for certain RegExp features are available.

slevithan:
@marcins Can you confirm whether you have this problem with XRegExp 3.0.0? You should not, since XRegExp 3.0.0 changed how support for the native /y flag is detected, to follow ES6.

(At the time of release, `RegExp.prototype.sticky` was not yet causing problems in any known browsers, and had been working in Firefox since the flag was introduced in 2008.)

marcins:
Apologies, you're right, turns out we were running off a "3.0.0-pre" commit hash from 2012. I bumped the version to 3.0.0 in our app and it appears to now work correctly in Canary with Experimental JS enabled.

slevithan:
Sweet. You should also get some other benefits with the upgrade including performance improvements. Props to @mathiasbynens for previously identifying and fixing the bug.

PR #102, opened 2015-09-15
Fix flag detection

mathiasbynens:
See the discussion over at https://github.com/slevithan/xregexp/commit/d93af4324e9384e457fb954eb1771520ba4654e5.

PR #123, opened 2015-12-17
More web-compatible way to feature-test for sticky RegExp support

vp2177:
This is to work around https://bugs.chromium.org/p/v8/issues/detail?id=4617

### DIFF
--- a/tests/perf/versions/xregexp-all-v1.5.1.js
+++ b/tests/perf/versions/xregexp-all-v1.5.1.js
@@ -119,7 +119,7 @@ if (XRegExp) {
             nativ.test.call(x, "");
             return !x.lastIndex;
         }(),
-        hasNativeY = RegExp.prototype.sticky !== undefined,
+        hasNativeY = RegExp.prototype.hasOwnProperty('sticky'),
         nativeTokens = {};
 
     // `nativeTokens` match native multicharacter metasequences only (including deprecated octals,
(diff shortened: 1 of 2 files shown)


mathiasbynens:
Thanks, @vp2177!

This is the same issue as https://github.com/slevithan/xregexp/commit/d93af4324e9384e457fb954eb1771520ba4654e5#commitcomment-13230259 (ref. #102, #121). Sorry for not spotting this particular instance before the v3 release :(

In case we want to avoid `hasOwnProperty`, the `try`-`catch` approach we’ve used before can be used.

vp2177:
The issue I see with `try-catch` is, in the latest V8 `RegExp.prototype.sticky` would throw, but that doesn't mean it's not supported (it is!).

mathiasbynens:
@vp2177 I meant like this: https://github.com/slevithan/xregexp/pull/102/files#diff-45ba574373e64db1c6c2653734c6e319R59

mathiasbynens:
Alternatively, to avoid both `RegExp.prototype` and `try`/`catch`:

```js
var emptyRegExp = /(?:)/;

var supportsSticky = 'sticky' in emptyRegExp;
var supportsUnicode = 'unicode' in emptyRegExp;
```

mathiasbynens:
Wait a second… This issue was fixed in #102 already, and XRegExp v3 includes this fix. This patch only changes the old versioned copies of XRegExp in the `test/perf` folder, which doesn’t really make sense.

If you see someone using an old version of XRegExp, please point them here and ask them to update to v3.

mathiasbynens:
old comments removed to avoid confusion

Issue #85, opened 2015-05-19
add \Q\E Escaping support


SReject:
As of current, block escaping([`\Q\E`](http://www.regular-expressions.info/characters.html#special)) is not supported and would be greatly appreciated if added. The usage allows for large blocks of text to be escaped without having to escape all 'dangerous' characters before compilation.

#### Rules:
Anything between `\Q` and the **first** encountered `\E` is seen as a string literal
`\Q` encountered within an open escape block should be interpreted as a literal "\Q"
`\E` is implied to be at the end of the pattern if not specified
`\\E` does **not** escape `\E` within escape blocks
`\E` encountered without a preceding open `\Q` should raise an error

#### Examples:

| Pattern | Matches | As |
| --- | --- | --- |
`\Qtext\E` | "text" | `/text/`
`\Q\Q\E` | "\Q" | /\\Q/
`\Qtext` | "text" | `/\Qtext\E/`
`\Qt.xt\E`  | "t.xt" | `/t\.xt/`
`\Q\\E` | "\" | `/\\/`
`text\E`| Erroneous

slevithan:
Thanks for the excellent writeup, @SReject!

I'm not currently planning to add this, because I think this syntax is little known/used, and has a safer alternative: manual use of `XRegExp.escape` and concatenation. This is safer because it avoids problems if "\E" appears within the text you are embedding.

That said, XRegExp is designed to support 3rd party syntax addons, and this one is easy to add if it is useful for you:

```js
XRegExp.addToken(
  /\\Q([\s\S]*?)(?:\\E|$)/,
  function (match) {
    return XRegExp.escape(match[1]);
  },
  {scope: 'all'}
);

// Test it
XRegExp('^\\Q(?*+)').test('(?*+)'); // -> true
XRegExp('^\\Q.\\E.').test('.x'); // -> true
XRegExp('^\\Q.\\E.').test('xx'); // -> false

// XRegExp already throws for escaped letters without native functionality
// unless they are overridden by addons
XRegExp('\\E'); // -> Uncaught SyntaxError: Invalid escape \E
```

mintern:
I implemented this support as an NPM module: <https://www.npmjs.com/package/xregexp-quotemeta>.

PR #137, opened 2016-06-03
Add quotemeta addon

mintern:
Java's [Pattern.escape](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-) and Perl's [quotemeta](http://perldoc.perl.org/functions/quotemeta.html) both allow a string delimited by `\Q..\E` to be interpreted literally, escaping all regex special characters. This addon implements quotemeta support for XRegExp.

### DIFF
--- a/src/addons/quotemeta.js
+++ b/src/addons/quotemeta.js
@@ -1,0 +1,29 @@
+/*!
+ * XRegExp Quote Meta 4.0
+ * <xregexp.com>
+ * Everlaw (c) 2016 MIT License
+ */
+
+module.exports = function(XRegExp) {
+    'use strict';
+
+    /**
+     * Adds support for \Q..\E escape sequences, used in Java and Perl regular expressions. All
+     * intervening characters are interpreted as literals instead of special regex characters.
+     *
+     * @requires XRegExp
+     */
+
+    XRegExp.addToken(
+        // Non-greedy matching is important for patterns that have multiple escaped sequences.
+        /\\Q((?!\\E).*?)\\E/,
+        function(match) {
+            return XRegExp.escape(match[1]);
+        },
+        {
+            scope: 'all',
+            leadChar: '\\'
+        }
+    );
+
+};
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ var XRegExp = require('./xregexp');
 
 require('./addons/build')(XRegExp);
 require('./addons/matchrecursive')(XRegExp);
+require('./addons/quotemeta')(XRegExp);
 require('./addons/unicode-base')(XRegExp);
 require('./addons/unicode-blocks')(XRegExp);
 require('./addons/unicode-categories')(XRegExp);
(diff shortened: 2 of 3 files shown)


slevithan:
Nicely done!

Closing this because I'm not planning to incorporate it in the main library, for reasons explained in #85. I've generally tried to be cautious about adding new syntax extensions.

mintern:
Oh, I didn't even see #85 before I implemented this.

Our use case is that we wish to share a regex for both Java server-side code and JavaScript. When `Pattern.escape` is called in Java (important for the same reasons that `XRegExp.escape` is useful), it inserts these `\Q..\E` characters. This addon allows that part of the regex to be interpreted.

If you don't wish for `quotemeta` to be included by default in `index.js`, would you be open to accepting only the first commit, so that `XRegExp` users can `require('quotemeta')` on an as-needed basis?

slevithan:
Yes, I'm open to that, but concerned about two precedents it would set:

1. Addons that aren't included in `xregexp-all.js`. If doing this, I'd probably want to drop 1 or 2 or maybe all of the existing addons from being included as well.
2. Optional syntax extensions. This would make the syntax of XRegExp unreliable/unpredictable when using XRegExp across projects. There would no longer be such a thing as "XRegExp 3" syntax, without qualifiers.

Number 2 is the more concerning, to me. Before enabling this, I'd want there to be a way for syntax extensions to be installable and uninstallable, and to be able to tell programmatically whether a syntax extension is enabled. An alternative model might be to allow subclassing XRegExp and to change syntax extensions in the process.

If you're interested in tackling something like that, that would be interesting to see. In the meantime, of course you should feel to use the `\Q\E` syntax extension in your own projects if it is useful to you. I'd recommend using the slight variation shown in #85 since it accounts for newlines and I'd expect the pattern used there to perform slightly better than repeating negative lookahead for each character.

mintern:
Both of your reasons make a lot of sense to me. I think it would be best suited as a separate `xregexp-quotemeta` library.

In addition to being more performant (as you mentioned), #85 is also much better specified: I wasn't aware that a terminal `\E` is unnecessary. With those ideas in mind, you pattern is much better IMO.

Would you prefer for me to create `mintern/xregexp-quotemeta`, or would it make more sense to host it as `slevithan/xregexp-quotemeta`, or something else?

slevithan:
`mintern/xregexp-quotemeta` sounds great. :)

mintern:
Just to be totally explicit, can you please confirm for me that you will license the code and tests in https://github.com/slevithan/xregexp/issues/85#issuecomment-108592346 under the same MIT License as this project? I will add `Steven Levithan (c) 2015 MIT License` to the license header of `quotemeta.js` and to the tests I plan to add, since that's the year of your PR comment.

slevithan:
Yes, all XRegExp code, tools, and tests are released under the MIT license, and any additional code I post in XRegExp issues or on XRegExp.com is released under the MIT license unless stated otherwise.

mintern:
Thank you very much!

mintern:
I implemented this support as an NPM module: <https://www.npmjs.com/package/xregexp-quotemeta>.

PR #143, opened 2016-08-07
follow-up on commit SHA-1: f5dbc86a9fb6657f8323d814534d9ae92436c8b7 (escaping literal brackets)

GerHobbelt:
follow-up on commit SHA-1: f5dbc86a9fb6657f8323d814534d9ae92436c8b7 (:: Correctly escape closing square brackets in two regular expression literals.) --> one more regex with an escaped = *literal* square bracket.

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -757,7 +757,7 @@ XRegExp.cache.flush = function(cacheName) {
  * // -> 'Escaped\?\ <\.>'
  */
 XRegExp.escape = function(str) {
-    return nativ.replace.call(toObject(str), /[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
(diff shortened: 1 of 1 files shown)


slevithan:
This isn't directly analogous to the related commit because you're escaping a literal square bracket *within* a character class. Is this avoiding any actual problem?

GerHobbelt:
No direct problem except now the code is not triggering my lint (ESLint+JSHint+JSCS) runs anymore; it is in line with https://github.com/slevithan/xregexp/commit/f5dbc86a9fb6657f8323d814534d9ae92436c8b7 as there Duktape was triggering a similar issue: in both situations the escaped brackets can be *implicitly* parsed as literals by the JS regex engine as they cannot form a regex `[...]` set, even when unescaped.
Escaping *all* meant-as-literal brackets can also help human confusing when 'reading' regexes.

PR #144, opened 2016-08-07
added API `XRegExp.isUnicodeSlug(name)` which checks whether a given `name` is a recognized Unicode Slug

GerHobbelt:
added API `XRegExp.isUnicodeSlug(name)` which checks whether a given `name` is a recognized Unicode Slug, e.g. `Number` (but not `Numeric` as `\p{Numeric}` is an _invalid_ Unicode `\p` regex slug.  This API is used by the GerHobbelt/jison app to help add Unicode support to the grammars being compiled. The way the API internals (**and documentation**, see the comment block in the patch) is written, i.e. that rather than a plain `true` boolean value, a 'truthy' Unicode slug definition is produced, is also used by the GerHobbelt/jison app: that code generator uses this artifact to greatly simplify lexer regex expansion and rewriting.

### DIFF
--- a/src/addons/unicode-base.js
+++ b/src/addons/unicode-base.js
@@ -225,19 +225,41 @@ module.exports = function(XRegExp) {
     };
 
     /**
-     * Test if the given name is a legal Unicode Slug for use in XRegExp `\p` or `\P` regex constructs.
+     * @ignore
+     *
+     * Return a reference to the internal Unicode definition structure for the given Unicode Property
+     * if the given name is a legal Unicode Property for use in XRegExp `\p` or `\P` regex constructs.
      *
      * @memberOf XRegExp
-     * @param {String} name Name by which the Unicode slug may be recognized (case-insensitive),
+     * @param {String} name Name by which the Unicode Property may be recognized (case-insensitive),
      *   e.g. `'N'` or `'Number'`.
-     *   
-     *   The given name is matched against all registered Unicode names and aliases.
      *
-     * @return {Object} Truthy when the name matches a Unicode slug (the internal slug definition
-     *   object is returned); `false` when the name does not match *any* Unicode name or alias. 
+     *   The given name is matched against all registered Unicode Properties and Property Aliases.
+     *
+     * @return {Object} Reference to definition structure when the name matches a Unicode Property;
+     * `false` when the name does not match *any* Unicode Property or Property Alias.
+     *
+     * @note
+     * For more info on Unicode Properties, see also http://unicode.org/reports/tr18/#Categories.
+     *
+     * @note
+     * This method is *not* part of the officially documented and published API and is meant 'for
+     * advanced use only' where userland code wishes to re-use the (large) internal Unicode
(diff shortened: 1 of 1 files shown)


src/addons/unicode-base.js
```diff
@@ -224,4 +224,20 @@ module.exports = function(XRegExp) {
...
+     * @memberOf XRegExp
+     * @param {String} name Name by which the Unicode slug may be recognized (case-insensitive),
+     *   e.g. `'N'` or `'Number'`.
+     *   
+     *   The given name is matched against all registered Unicode names and aliases.
+     *
+     * @return {Object} Truthy when the name matches a Unicode slug (the internal slug definition
+     *   object is returned); `false` when the name does not match *any* Unicode name or alias. 
+     */
+    XRegExp.isUnicodeSlug = function(name) {
```
slevithan: Variables/args named `slug` in Unicode Base have always *already* had their value passed through `normalize`. In other words it's lowercase and has had any hyphens, spaces, and underscores removed. It's a unique value that represents all the valid variations of a particular Unicode property name (not including explicit aliases). Therefore IMO the word "slug" isn't appropriate in this method's name.

Also, I'm not a fan of this API that returns a truthy or boolean value. Based on the name this should always return a boolean. If you need to get back the data for the Unicode property, that's fine, but I don't think this should be an official/documented method. Recommend changing the method to something like:

```js
// Intentionally undocumented
XRegExp._getUnicodeData = function(name) {
  var slug = normalize(name);
  return unicode[slug];
}
```

GerHobbelt:
Yep, the API is a little nasty that way; I like the suggested alternative.

GerHobbelt:
About the 'intentionally undocumented': I like to have the comment/doc in there, but could tag is as an `\@ignore` section, so only folks like me who RTFM will find the otherwise 'private' API -- which is *not* 'private' as I use it in my jison code, so it *is* an API, just one we don't want advertised in the documentation as it's a little hairy: it (intentionally) exports internal storage structures.

See also the JSDoc docs for `\@ignore`: http://usejsdoc.org/tags-ignore.html

GerHobbelt:
Code updated: 
Adjust the code to rename the API to `_getUnicodeProperty` -- I also went and checked the official term for these 'slugs': as I intend them, they're called 'categories' or 'Unicode Properties' (http://unicode.org/reports/tr18/#Categories)
--> hence `_getUnicodeProperty` instead of your suggested `_getUnicodeName`  ;-)

slevithan:
Thanks for the updates, @GerHobbelt! Before I merge, can you make a few more changes please:

* Remove the `|| false`, so this just returns `undefined` if no data is available.
* Add basic tests.
* Run `npm run build` to update `xregexp-all.js`. This will add changes from your other recent pull requests that have been merged, which is fine.
* Squash the commits into one.

GerHobbelt:
WILCO. Will rebase this to latest master when time available.

GerHobbelt:
🛃  *ping*?

PR #145, opened 2016-08-07
migrated performance tests to use latest BenchmarkJS v2.1.1

GerHobbelt:
migrated performance tests to use latest BenchmarkJS v2.1.1 (which requires lodash and optional platform libraries to be loaded as well: these are included in the tests/vendor/benchmark.js/ directory)

### DIFF
--- a/tests/perf/index.html
+++ b/tests/perf/index.html
@@ -45,6 +45,8 @@
         }());
     </script>
 
+    <script src="../vendor/benchmark.js/lodash.js"></script>
+    <script src="../vendor/benchmark.js/platform.js"></script>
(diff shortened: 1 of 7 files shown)


slevithan:
Thanks for putting in the effort to upgrade this!

Are there any tradeoffs to upgrading? Is the resulting behavior exactly the same? If not, what changes with your diff?

GerHobbelt:
AFAICT nothing changes in the behaviour (that was the intent, anyway); it's just that I've been working with the latest BenchmarkJS code and use xregexp extensively in my toolchain - when I observed that xregexp used an old BenchmarkJS version itself and thought it be handy to test both at the same time in the same manner, hence using the same performance testing framework.

PR #153, opened 2017-02-07
Optimize sticky matching

pygy:
[Baseline (build of the current master)](https://github.com/pygy/xregexp/blob/c28faa33c5be869e68e626da27021d35d1a02793/tests/perf/index.html)

[Proposed patch](https://cdn.rawgit.com/pygy/xregexp/8f8acf728ec840e35868b5a62e54878ac99a7465/tests/perf/index.html)

[Test suite post-build](https://cdn.rawgit.com/pygy/xregexp/8f8acf728ec840e35868b5a62e54878ac99a7465/tests/index.html)

I've only tested this in an old Chromium build so far that doesn't have `y`. The sticky benchmark performance is one order of magnitude better.

I've also had 2x improvements in the non-cache/non-flush XRegExp constructor tests, but they are not reliably reproducible.

I'll try it in SauceLabs later today for older IE versions.

Edit: one order of magnitude, not "orders"... Still, better ;-). Sticky matching goes from `O(N * complexity(regexp))` to `O(1 * complexity(regexp))` where N is the string length.

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -203,7 +203,7 @@ function copyRegex(regex, options) {
     // unnecessary for regexes constructed by `XRegExp` because the regex has already undergone the
     // translation to native regex syntax
     regex = augment(
-        new RegExp(regex.source, flags),
+        new RegExp(options.source || regex.source, flags),
         hasNamedCapture(regex) ? xData.captureNames.slice(0) : null,
         xregexpSource,
         xregexpFlags,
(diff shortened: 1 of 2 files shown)


src/xregexp.js
```diff
@@ -791,21 +793,28 @@ XRegExp.escape = function(str) {
...
+        fakeY = false,
         match,
         r2;
 
     addY = hasNativeY && !!(sticky || (regex.sticky && sticky !== false));
     if (addY) {
         cacheKey += 'y';
+    } else if (sticky) {
+        cacheKey += 'fakeY'
+        fakeY = true
```
mathiasbynens: nit: missing semicolons (please follow the existing coding style)

pygy: Oh, yeah, sorry, I don't use them in my own code and forgot to check what you're using. I'll push a rebased, fixed version.

pygy: BTW, I'm not entirely satisfied by the details of this patch. Having a FakeY-specific pathway in `copyRegex` is suboptimal... Maybe we could instead honour `options.source`. `XRegExp.exec` could then inject `source: regex.source + '|()'`...

pygy:
Ok, done, with a generic `options.source` for `copyRegex` in order to keep the `FakeY`-specific logic in `XRegExp.exec`, and I've added comments that describe the trick.

The PR UI shows the commits ordered by date/time rather than [commit order](https://github.com/pygy/xregexp/commits/fast-fake-y). I added the build-before-the-patch afterwards and rebased, I didn't expect GH to show the commits out of log order...

There are similar perf gains in IE9 (from 60,000 to 450,000 ops/sec in sticky mode).

slevithan:
This is really nice. Simpler and more self-contained than I expected this trick would be. Thank you!

pygy:
The infrastructure was there, I just needed to tap into it...

I just realised that I was allocating `sourceY` even when `r2` was already cached... Update incoming (with the `lastIndex` patch). Do you want builds with each modification?

slevithan:
Yes, please. No preference on whether builds are included with modifications or in a follow up diff. Just hoping to always keep the build up to date.

Issue #152, opened 2017-02-05
Proof of concept: simulating sticky mode in pre-ES6 environemnts...


pygy:
It is possible, with some hand holding, to simulate sticky matching with normal RegExps.

```JS
function stickify(original) {
  var matcher = new RegExp(original.source+'|()', 'g'+original.flags) //imagine we are deduping `g`s here
  return {
    exec: function(str) {
      matcher.lastIndex = this.lastIndex;
      var res = matcher.exec(str)
      if (res && res.pop() == null) return res // setting `this.lastIndex`  elided for the demo
      else return null
    },
    lastIndex: 0
  }
}
```

By appending `|()` to the source, you ensure that the match will succeed no matter what at the given index. If `original` doesn't match, the last capture will be `''`, `null` otherwise.

The trouble now is to convince UAs to treat the resulting object as a RegExp in `String.prototype.*` context... I suppose that in ES6 you could extend native RegExps, but the polyfill is not needed there since ES6 introduces the `y` flag...

Edit: now the code actualy works :)

slevithan:
XRegExp already supports sticky mode for ES3+ using `XRegExp.exec`, `XRegExp.test`, and `XRegExp.matchRecursive`. I've long wanted to extend this to more methods, but have avoided this pending a broader API change (see #109).

pygy:
Ooh, that's nice, I had missed the sticky mode... IIUC with the current implementation, in sticky mode, the regexp still searches till the end of the subject and fails if it succeeds at a position different from `lastIndex`... 

The snippet above doesn't search beyond the current index, but you have to create an extra RegExp object... It may be cached in the `REGEX_DATA` field though so it would be a one time cost.

Interrested in a PR?

slevithan:
[...]

PR #154, opened 2017-02-12
Don't allocate the sticky regex source when the resulting regexp is already cached

pygy:
As discussed previously in #152 and #153

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -815,7 +815,7 @@ XRegExp.escape = function(str) {
 XRegExp.exec = function(str, regex, pos, sticky) {
     var cacheKey = 'g',
         addY = false,
-        sourceY,
+        fakeY = false,
         match,
         r2;
 
@@ -828,7 +828,7 @@ XRegExp.exec = function(str, regex, pos, sticky) {
         // and will not search the rest of the subject string. We'll know that the original regex
         // has failed if that last capture is `''` rather than `undefined` (i.e., if that last
         // capture participated in the match).
-        sourceY = regex.source + '|()';
+        fakeY = true;
         cacheKey += 'FakeY';
     }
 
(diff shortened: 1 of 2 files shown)


src/xregexp.js
```diff
@@ -839,7 +839,7 @@ XRegExp.exec = function(str, regex, pos, sticky) {
         regex[REGEX_DATA][cacheKey] = copyRegex(regex, {
             addG: true,
             addY: addY,
-            source: sourceY,
+            source: fakeY? regex.source + '|()' : void 0,
```
slevithan: Nits: Space before `?`, and please just use `undefined` rather than `void 0`.

pygy:
Thinking more about this, I'm not sure what to do for regexps that are both global and sticky. It makes little sense to set both flags, but it is legal AFAIK... Scaling down.

src/xregexp.js
pygy: Sure, updating

slevithan:
To be clear, the "fast sticky polyfill" rev is just avoiding an unnecessary string concat in some cases, and otherwise not changing behavior at all, correct?

pygy:
Correct. I've removed the flags thing, I'll make another PR once the `gy` situation is more clear.

slevithan:
> I've removed the lastIndex thing, I'll make another PR once the `gy` situation is more clear.

Thanks--perfect! You removing it from this PR made me more confident to merge the other change. The `gy` situation was still a bit fuzzy to me even though you included some great tests (thank you!), so I was hoping to come back to that to examine more fully. BTW, the reason XRegExp isn't yet fully tackling `lastIndex` setting for `/y` is because most of the related code was written before the existence of ES6 `/y`. It was based on `/y` in old versions of Firefox. I appreciate you tackling this somewhat confusing area to bring XRegExp up to spec with ES6 in handling of `/g`, `/y`, and `/gy`.

slevithan:
@pygy BTW, if you submit robust `/y` + `lastIndex` handling, I'd be happy to publish a v3.1.2 that gets your `/y` speed and correctness updates out there sooner.

pygy:
Ok, According to both the [latest draft](https://tc39.github.io/ecma262/#sec-regexpbuiltinexec) and the [ES6](http://www.ecma-international.org/ecma-262/6.0/#sec-regexpbuiltinexec) spec, `y` takes precedence over `g`. I'll look into it tomorrow evening.

PR #156, opened 2017-02-14
Add max_line_length to .editorconfig

josephfrazier:
It looks like comments have manually been wrapped to 100 characters, so
use that as the [max_line_length] in the .editorconfig file.

[max_line_length]: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length


slevithan:
Thanks.

PR #157, opened 2017-02-14
Allow using positive bare integers as capture names with explicit capture

josephfrazier:
This makes it easier to do transformations like the following without
breaking code that assumes `$1` can be used in a replacement:
From: `/(?:not captured) (captured)/`
To: `XRegExp('(?n)(not captured) (?<1>captured)')`

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -1552,9 +1552,7 @@ fixed.replace = function(search, replacement) {
                     // 1. Backreference to numbered capture, if `n` is an integer. Use `0` for the
                     //    entire match. Any number of leading zeros may be used.
                     // 2. Backreference to named capture `n`, if it exists and is not an integer
-                    //    overridden by numbered capture. In practice, this does not overlap with
-                    //    numbered capture since XRegExp does not allow named capture to use a bare
-                    //    integer as the name.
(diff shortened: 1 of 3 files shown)


slevithan:
This is intentionally disallowed as it creates ambiguity and confusion. Why allow the confusion of something like `(?<2>.)(?<1>.)` if you don't have to? Also IMO creating a special case rule about allowed group names based on the presence of a flag is needlessly complicated/confusing.

I'm not seeing the value of `(?n)(not captured) (?<1>captured)` vs `(?:not captured) (captured)` or `(?n)(not captured) (?<$1>captured)`. Even though the latter would need to be referenced differently outside of the regex, at least it makes it clear that this is an explicit name rather than an auto-assigned number.

In general, XRegExp regex and replacement text syntax prefers strictness and helping people avoid shooting themselves in the foot (e.g. by disallowing octals, etc.).

josephfrazier:
> I'd recommend ... `(?n)(not captured) (?<$1>captured)`

Oops, I had completely overlooked that you can still use numbered references to the named groups >.< That completely solves my problem! I definitely see how this change would cause confusion. Thanks for the quick response.

PR #161, opened 2017-02-16
Also update .lastIndex on natively sticky regexps

pygy:
As mentioned in https://github.com/slevithan/xregexp/pull/154#issuecomment-280164443 regexps that have both the `global` and `sticky` flag ignore the former and do sticky matches. So there's nothing special to add to their handling AFAICT.

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -857,7 +857,7 @@ XRegExp.exec = function(str, regex, pos, sticky) {
         match = null;
     }
 
-    if (regex.global) {
+    if (regex.global || regex.sticky && sticky !== false) {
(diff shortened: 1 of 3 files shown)


pygy:
Wait, there's more... I haven't updated `fixed.exec` (and I wonder how the tests actually pass).

slevithan:
Hey, no rush! And thanks. :D

slevithan:
@pygy, are you still expecting to update this?

slevithan:
In case it's helpful, <http://2ality.com/2015/07/regexp-es6.html> describes the rules for `lastIndex` with `/y` in detail.

axljoker03 (approved):

Issue #162, opened 2017-02-18
(Firefox) XRegExp.build with 'x' flag doesn't ignore newlines


josephfrazier:
In Firefox, `XRegExp.build("\n", {}, 'x')` returns `/\n/`, whereas Chrome correctly returns `/(?:)/`. Here's a JSFiddle demo: https://jsfiddle.net/zz4gj26r/

However, if we make the 'x' flag a leading mode modifier instead, Firefox now correctly returns `/(?:)/`. Here's a JSFiddle for that as well: https://jsfiddle.net/nz0rtt0y/

For what it's worth, PhantomJS seems to have the same behavior as Firefox.

---

I'll probably keep looking into this, but I thought I'd go ahead and make an issue for it.

PR #163, opened 2017-02-18
Pass XRegExp.build flags through asXRegExp to XRegExp constructor

josephfrazier:
This fixes https://github.com/slevithan/xregexp/issues/162 by making
sure that all the flags passed to `XRegExp.build` make it to the
`XRegExp()` call, rather than just the flags used as leading mode
modifiers in the pattern.

### DIFF
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -85,6 +85,12 @@ module.exports = function(XRegExp) {
      */
     XRegExp.build = function(pattern, subs, flags) {
         var inlineFlags = /^\(\?([\w$]+)\)/.exec(pattern),
+            // These flags will be passed to the asXRegExp calls for `pattern` and for every
+            // subpattern in `subs`. This is to work around the following browser bugs:
+            //
+            // * Firefox converts '\n' to a regex that contains the literal characters \ and n
+            //   You can verify this by running `console.log(RegExp('\n').source)`
+            //   See here for more details: https://github.com/slevithan/xregexp/pull/163
             asXRegExpFlags = (flags || '').indexOf('x') > -1 ? 'x' : '',
             data = {},
             numCaps = 0, // 'Caps' is short for captures
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -38,22 +38,24 @@ module.exports = function(XRegExp) {
     }
 
     /**
-     * Converts the provided value to an XRegExp. Native RegExp flags are not preserved.
+     * Converts the provided value to an XRegExp. Native RegExp flags are not preserved. Flags can
+     * be provided as a separate argument.
      *
      * @private
      * @param {String|RegExp} value Value to convert.
+     * @param {String} [flags] Any combination of XRegExp flags.
      * @returns {RegExp} XRegExp object with XRegExp syntax applied.
      */
-    function asXRegExp(value) {
+    function asXRegExp(value, flags) {
         return XRegExp.isRegExp(value) ?
             (value[REGEX_DATA] && value[REGEX_DATA].captureNames ?
                 // Don't recompile, to preserve capture names
                 value :
                 // Recompile as XRegExp
-                XRegExp(value.source)
+                XRegExp(value.source, flags)
             ) :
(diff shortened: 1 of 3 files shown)


slevithan:
I'm not sure this change is appropriate. The flags are already passed to the XRegExp constructor correctly at <https://github.com/slevithan/xregexp/blob/ce6dc9a13bc1b87f1bd2f2c77f7a686d0c1742d6/src/addons/build.js#L186>. You can see that the `x` flag is being applied even without this change via something like `XRegExp.build(" ", {}, 'x')` or `XRegExp.build("   \n", {}, 'x')` instead of `XRegExp.build("\n", {}, 'x')`. The problem is just with the use of `\n`.

This seems to be a quirk in Firefox (not certain it's technically a big). What's happening is that when the string pattern is passed through `asXRegExp` and converted into a regex (which is done to throw errors on octals, ensure the outer pattern is independently valid, and help keep things simple), Firefox is changing the string `'\n'` (LF, char code 10) to the regex pattern (`/\n/`), which seems inappropriate to me even if it's not technically a bug (not sure what the spec might have to say about this edge case which only matters when looking at the string `source` of the resulting regex). The string source of this regex now contains the literal characters `\` and `n` (codes 92 and 110), rather than LF. :-/ So then when this string is passed to the `XRegExp` constructor, the `x` flag correctly doesn't apply to it anymore.

In order to fix this, either Firefox needs to change it's generated regex source for literal LFs, or `XRegExp.build` needs to apply some special processing to avoid this case. That special handling might need to be applying the `x` flag in calls to `asXRegExp` (though not necessarily applying any other flags), but in that case this should also happen for the calls to `asXRegExp` for the subpatterns, not just the outer pattern. The same bug is undoubtedly also happening for the subpatterns in XRegExp.build's second argument.

Applying all regex flags as part of `asXRegExp` (as done in this diff, and as opposed to just applying `x` when appropriate in `asXRegExp`) might be the right way to go in order to keep things simple. But there might be side effects in edge cases, perhaps with weird user-added custom flags that shouldn't be applied until the full XRegExp.build pattern is constructed. I'm not sure. :-/

josephfrazier:
Thanks for looking into this so thoroughly!  Your comment really helps explain what exactly is going on here.

I see why the changes I've made so far aren't quite right, due to the `x` flag not applying to subpatterns, as well as the potential issue with applying other flags across the board. I'll see if I can fix it to only apply the `x` flag, and to apply it to all subpatterns as well. Would you be more comfortable with merging the changes then?

---

For context, here's an [example workaround](https://github.com/OctoLinker/browser-extension/blob/11dec0e0b16e1f09d00b9df67faca2f36bdccd60/packages/helper-grammar-regex-collection/regex.js#L19-L21) for https://github.com/slevithan/xregexp/issues/162, so that code like [this](https://github.com/OctoLinker/browser-extension/blob/11dec0e0b16e1f09d00b9df67faca2f36bdccd60/packages/helper-grammar-regex-collection/index.js#L14-L18) works correctly.

slevithan:
> I'll see if I can fix it to only apply the `x` flag, and to apply it to all subpatterns as well. Would you be more comfortable with merging the changes then?

Since there's a real bug here, yes I'd be happy to merge in a fix like you described, especially if it comes with a comment explaining why this workaround is necessary.

josephfrazier:
Thanks, I've added an explanatory comment. Let me know if you'd like me to change the wording, or rearrange/squash the commits before this can be merged.

josephfrazier:
Thanks for the merge! For what it's worth, I reported a Firefox bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1350575

slevithan:
Thanks for reporting the bug and providing the fix!

josephfrazier:
Sure thing! After a bit more investigation, it seems like Firefox is actually correct here (see https://github.com/kangax/compat-table/issues/798), but I think it's still appropriate to keep this change in, since it helps keep XRegExp behavior intuitive and consistent across browsers.

PR #164, opened 2017-02-19
Cleanup (?:) from beginning/end of groups

josephfrazier:
This simplifies the compiled expressions by ensuring that groups don't
start or end with `(?:)`. For instance, this code:

```js
// Using named capture and flag x (free-spacing and line comments)
date = XRegExp('(?<year>  [0-9]{4} ) -?  # year  \n' +
               '(?<month> [0-9]{2} ) -?  # month \n' +
               '(?<day>   [0-9]{2} )     # day     ', 'x');
```

now compiles to this pattern:

```regex
([0-9]{4})(?:)-?(?:)([0-9]{2})(?:)-?(?:)([0-9]{2})(?:)
```

instead of

```regex
((?:)[0-9]{4}(?:))(?:)-?(?:)((?:)[0-9]{2}(?:))(?:)-?(?:)((?:)[0-9]{2}(?:))(?:)
```

Here are the two patterns side by side, with whitespace inserted into the new one to illustrate the differences:

```regex
(    [0-9]{4}    )(?:)-?(?:)(    [0-9]{2}    )(?:)-?(?:)(    [0-9]{2}    )(?:)
((?:)[0-9]{4}(?:))(?:)-?(?:)((?:)[0-9]{2}(?:))(?:)-?(?:)((?:)[0-9]{2}(?:))(?:)
```

### DIFF
--- a/tests/spec/s-xregexp.js
+++ b/tests/spec/s-xregexp.js
@@ -182,6 +182,19 @@ describe('XRegExp()', function() {
         expect(function() {XRegExp('(\\(?:)');}).not.toThrow();
     });
 
+    it('should not put (?:) at beginning/end of groups', function() {
+        var regex = XRegExp('( [0-9]{4} ) -?  # year  \n' +
+                            '( [0-9]{2} ) -?  # month \n' +
+                            '( [0-9]{2} )     # day     ', 'x');
+        expect(regex.source).toEqual('([0-9]{4})(?:)-?(?:)([0-9]{2})(?:)-?(?:)([0-9]{2})(?:)');
+    });
+
+    it('should leave escaped (?:) at end of group', function() {
+        var regex = XRegExp('(.(\\(?:))');
+        expect(regex.test('x:')).toBe(true);
+        expect(regex.test('x(:')).toBe(true);
+    });
+
     it('should store named capture data on regex instances', function() {
         // The `captureNames` property is undocumented, so this is technically just testing
         // implementation details. However, any changes to this need to be very intentional
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -332,7 +332,7 @@ function isPatternNext(pattern, pos, flags, needlePattern) {
         // Ignore any leading inline comments
         [inlineCommentPattern];
     return nativ.test.call(
-        RegExp('^(?:' + patternsToIgnore.join('|') + ')*(?:' + needlePattern + ')'),
+        new RegExp('^(?:' + patternsToIgnore.join('|') + ')*(?:' + needlePattern + ')'),
         pattern.slice(pos)
     );
 }
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -1775,7 +1775,10 @@ XRegExp.addToken(
     function(match, scope, flags) {
         // Keep tokens separated unless the following token is a quantifier. This avoids e.g.
         // inadvertedly changing `\1(?#)1` to `\11`.
-        return isQuantifierNext(match.input, match.index + match[0].length, flags) ?
+        return (
+            match.input.charAt(match.index - 1) === '(' ||
+            isQuantifierNext(match.input, match.index + match[0].length, flags) ||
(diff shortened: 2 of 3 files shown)


slevithan:
It looks like this replaces a few patterns:

* `(?:))` -> `)`
* `((?:)` -> `(`
* `(?:(?:)` -> `(?:`

In addition to `(?:)(?:)` -> `(?:)`, which is already done prior to these changes.

I agree with these improvements in principle but have two concerns in practice: impact on perf, and correctness.

For perf, what is the effect seen in `/tests/perf/index.html`, specifically on the "Constructor with short pattern" for the "XRegExp with pattern cache flush" test?

As for correctness, it's possible to break this in cases like `(.(\(?:))`, which should match e.g. `'x:'`, but would be converted to `(.(\)` and produce a syntax error. Avoiding this problem would likely lead to greater perf/complexity impact.

slevithan:
An alternative strategy for keeping generated regexes clean might be to give token handler functions access to the preceding XRegExp token so they can more smartly return `(?:)` only when really needed. (Search `xregexp.js` for the 3 cases of `'' : '(?:)'` to see where this is used.)

josephfrazier:
Thanks for the thorough review! I see that my original solution wasn't correct, with the `(.(\(?:))` case you pointed out, and I agree that it'd be better to avoid inserting the extra `(?:)` in the first place, rather than removing them after the fact.

Accordingly, I've force-pushed a different set of commits into this branch that uses the alternative strategy, and additionally tests the `(.(\(?:))` case for good measure. This required some refactoring before the meaningful changes, and I also did a small refactoring after, so it's probably easiest to review the commits one-by-one. I tried to make the commit messages informative as well. Let me know what you think!

slevithan:
Thanks! I still need to look over the new set of diffs closely, but I love the idea of no longer inserting `(?:)` where it isn't needed.

Note to self: See if https://github.com/slevithan/xregexp/blob/a5e5f43a65c17547cb7be7c2859d1022f2bcbc7f/src/addons/build.js#L25-L28 is still needed after these changes.

tests/spec/s-xregexp.js
```diff
@@ -182,6 +182,19 @@ describe('XRegExp()', function() {
         expect(function() {XRegExp('(\\(?:)');}).not.toThrow();
     });
 
+    it('should not put (?:) at beginning/end of groups', function() {
+        var regex = XRegExp('( [0-9]{4} ) -?  # year  \n' +
+                            '( [0-9]{2} ) -?  # month \n' +
+                            '( [0-9]{2} )     # day     ', 'x');
+        expect(regex.source).toEqual('([0-9]{4})(?:)-?(?:)([0-9]{2})(?:)-?(?:)([0-9]{2})(?:)');
```
slevithan: This is enforcing the inclusion of multiple `(?:)` empty groups that aren't needed for this regex to operate correctly. The test should be re-framed to not enforce anything that is unnecessary.

josephfrazier: Oops, yeah it is a bit brittle. We could change it to reject certain substrings, but then we might end up duplicating some of the logic in the non-test code. What if we changed it to something a little simpler, but still future-proof, like this?

```js
expect(regex.source.length <= 54).toBe(true); // 54 is the length of the current result
```

slevithan:
Testing string length wouldn't verify that it's working. I've gone ahead and updated this in 622aaf37d71298d2e0b25f90318fbeed303d38fc.

The "Constructor with x flag, whitespace, and comments" perf test is now meaningfully slower than in 3.1.1, as a result of these changes. You could easily create examples that are even more affected, since any regex syntax that triggers the new code will be slower with every invocation. I'll probably try to look at speeding this back up later, both by avoiding the string concatenation in `isPatternNext` (possibly going back to regex literals and making this specific to quantifiers, even though the current code is definitely more readable) and by updating `runTokens` to make already processed tokens available to token handler functions. This would make custom XRegExp tokens more powerful in general and also allow adding support for more cases where `(?:)` shouldn't be inserted (e.g. just after an opening `(?:`) without further slowing things down.

PR #167, opened 2017-03-30
Include build package in package.json for npm installs.

### LINKED ISSUE #139: After pull can not see xregexp-all.js
After I pulled your code. I don't see xregexp-all.js. Event I run with "slevithan/xregexp#63273f93d97dd7345b6cc57b59ae5d6213d29b8d". But I saw in github repository. So I there are any way to use 'xregexp-all.js' file without run build script?

Thanks you for help!

kaidjohnson:
While `xregexp-all.js` exists in the repo, it isn't included in the npm install. Adding the file to the package.json ensures npm installs provide a production-ready version for non-node applications.

This PR fixes #139

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,10 +15,11 @@
     "regular expression",
     "unicode"
   ],
-  "main": "./src/index.js",
+  "main": "xregexp-all.js",
   "files": [
     "src",
-    "xregexp-all.js"
+    "xregexp-all.js",
+    "LICENSE"
   ],
   "scripts": {
     "build": "browserify src/index.js --standalone XRegExp > xregexp-all.js",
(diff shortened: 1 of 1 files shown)


slevithan:
Thanks!

Can you clarify about removing `src` from `files`? Why is that necessary, and will the `src` directory still be included so that people can `require` individual modules?

kaidjohnson:
Yeah, I missed that rather crucial bit about supporting the modules therein. Sorry about that!

I've fixed this issue by updating the `main` to point to the built file now to ensure `require('xregexp');` pulls in the complete package. The `src` files are still included for incremental builds. Added the LICENSE for npm installs as well, per best practices.

Issue #106, opened 2015-09-26
Add `conjunction` option for XRegExp.union and allow setting to 'none'


slevithan:
`XRegExp.union` implicitly uses "or" as the conjunction for the strings and regexes it merges. However, "none" would also be useful in some cases (verified by #43). Backreferences would be safely rewritten regardless of the provided conjunction.

```
var regexes = [/man/, /bear/, /pig/];

// Current handling: absent conjunction is implicit 'or'
XRegExp.union(regexes, 'i');
// -> /man|bear|pig/i

// New feature: conjunction can explicitly be set to 'or'
XRegExp.union(regexes, 'i', {conjunction: 'or'});
// -> /man|bear|pig/i

// New feature: conjunction can be set to 'none'
XRegExp.union(regexes, 'i', {conjunction: 'none'});
// -> /manbearpig/i
```

Offering a "none" conjunction option will resolve #43. It's also the only new handling currently planned for support via `conjunction`, unless people request additional options. That's because `none` is likely the most useful alternative and is trivial to implement, unlike other options. However, `and`, `xor`, and various other options are possible and may be added later.

```
`none`
/manbearbig/

`or`
/man|bear|pig/

`xor`
/(?!=bear|pig)man|(?!man|pig)bear|(?!man|bear)pig/

...or to ensure backreferences are ordered correctly:
/(?=man|bear|pig)(?:(?!=bear|pig)man|(?!man|pig)bear|(?!man|bear)pig)/

`and`
/(?=bear)(?=pig)man/

...or to ensure backreferences are ordered correctly:
1. /(?=man)(?=bear)pig/ <-- this changes what's matched
2. /(?=man)(?=bear)(?=pig)man/

Additional exotic conjunction option outputs.
Not sure what these would be named.
Also it's generally better to solve these problems using multiple regexes.
and-within-string: /(?=[\S\s]*?man)(?=[\S\s]*?bear)(?=[\S\s]*?pig)(?:man|bear|pig)/
xor-within-string: /(?!=[\S\s]*?bear|[\S\s]*?pig)man|(?!=[\S\s]*?man|[\S\s]*?pig)bear|(?!=[\S\s]*?man|[\S\s]*?bear)pig/
```

slevithan:
Should also check whether this change can be taken advantage of to simplify `XRegExp.build`, which uses some similar logic. `XRegExp.union` is also used in `XRegExp.matchRecursive` for the `escapeChar` option.

PR #169, opened 2017-04-08
Add `conjunction` option for XRegExp.union and allow setting to 'none'

josephfrazier:
The `XRegExp.union` calls in the `build` and `matchRecursive` addons are
also updated to explicitly specify `{conjunction: 'or'}`.

Fixes https://github.com/slevithan/xregexp/issues/106

### DIFF
--- a/xregexp-all.js
+++ b/xregexp-all.js
@@ -11,7 +11,9 @@ module.exports = function(XRegExp) {
 
     var REGEX_DATA = 'xregexp';
     var subParts = /(\()(?!\?)|\\([1-9]\d*)|\\[\s\S]|\[(?:[^\\\]]|\\[\s\S])*\]/g;
-    var parts = XRegExp.union([/\({{([\w$]+)}}\)|{{([\w$]+)}}/, subParts], 'g');
+    var parts = XRegExp.union([/\({{([\w$]+)}}\)|{{([\w$]+)}}/, subParts], 'g', {
+        conjunction: 'or'
+    });
 
     /**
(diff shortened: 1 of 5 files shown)


xregexp-all.js
```diff
@@ -298,7 +300,7 @@ module.exports = function(XRegExp) {
             // Using `XRegExp.union` safely rewrites backreferences in `left` and `right`
             esc = new RegExp(
                 '(?:' + escapeChar + '[\\S\\s]|(?:(?!' +
-                    XRegExp.union([left, right]).source +
+                    XRegExp.union([left, right], basicFlags, {conjunction: 'or'}).source +
```
slevithan: It's not necessary and possibly inappropriate to add `basicFlags` here, since those flags were already applied to `left` and `right` when those strings were passed through the `XRegExp` constructor a few lines up. In most cases it wouldn't matter if you apply the flags again, but in edge cases with user-added custom flag addons that transform syntax that they also output, this could lead to transforming the original output in a way that isn't intended.

I noted it's "possibly" inappropriate, because honestly this edge case is unavoidable given the flexibly of user-added syntax transforms. They could be transforming their own output even in cases that don't rely on a flag. Hopefully no addons do that.

I'll go ahead and merge this though, since I can revert this part while adding a comment about it.

josephfrazier: Oops, I wasn't sure which flags to use, so I went with what was already used elsewhere. However, I definitely didn't consider the implications of it as thoroughly as you have, and I see now that the empty list of flags is the way to go. Thanks for merging anyway and fixing things up after the fact!

Issue #107, opened 2015-09-26
Code cleanup and style updates


slevithan:
- <s>Consider renaming local variable `self` as `XRegExp`.</s> Completed in #114 by @d10.
- New `var` on each line, and don’t always declare vars at top of function.
- <s>Make function comments line up with the code below them rather than starting at char 0.</s>
- Consider switching from 4- to 2-space tabs.
- More named temporary variables, less variable reuse.
- Change `<li>`s to asterisks or hyphens in code comments.
- Update comments in code to use e.g. `"flag `g`"` and `"flags `g` or `y`"` instead of `"flag g"` and `"flags gy"`.
- Escape literal `"{"` chars in regexes, since it's technically required. See `nativeTokens`, `replacementToken`, the letter escape and `\u{N..}` tokens, `XRegExp.build`, etc.
- <s>Remove protection against overriding `undefined`. It's immutable in ES5 anyway.</s>
- <s>Expose `_dec`, `_hex`, and `_pad4` on XRegExp so they don’t need to be duplicated by Unicode Base.</s>
- <s>Remove `@memberOf`, `@private`, and</s> `@static` <s>params.</s>

PR #173, opened 2017-04-15
Change `<li>`s to hyphens in code comments

josephfrazier:
As seen in https://github.com/slevithan/xregexp/issues/107:

> * Change `<li>`s to asterisks or hyphens in code comments.

I used hyphens rather than asterisks because each comment line already
starts with an asterisk, so hyphens visually stand out more.


slevithan:
Thanks!

Issue #103, opened 2015-09-15
Implement tagged template string


teohhanhui:
As per your gist from 2012: https://gist.github.com/slevithan/4222600

slevithan:
IMO this is not a priority until more browsers support tagged template strings natively.

Some changes in thinking since writing that gist:

* Not sure what the name of the tag should be, but probably not `r`. Maybe `XRegExp.make` or `XRegExp.create`. You could do <code>XRegExp.make(flags)\`...\`</code> to provide flags.
* Probably would want this to accept regexes as raw strings, but it make sense to support atomic regex/string interpolation (but not support flags for interpolated regexes).
* Would want to implicitly enable free-spacing, but also maybe implicitly enable "explicit capture" and ES6's unicode flag. Probably don't want to offer multiple tags with any variation of implicit flags.
* Would want to add `$<n>` and `$<name>` as alternative backreference syntax for `${n}` and `${name}` in replacement text, so they can be used in ES6 template strings. This would be backward compatible, since `$<` is currently an error in XRegExp replacement text.

teohhanhui:
Do tagged template strings not work with a transpiler like Babel?

slevithan:
I actually don't know whether it supports creation of arbitrary new tag functions. Of course, you can and should use Babel today to get raw strings and multiline strings. No XRegExp changes are needed for that.

rauschma:
@slevithan A tagged template literal is just another way of making a function call: the first parameter is an Array with the static parts of the template literal, the remaining parameters are the dynamic `${}` parts of the template literal.

Thus, you can even write tag functions in ES5.

More info: http://exploringjs.com/es6/ch_template-literals.html#_implementing-tag-functions

Kagami:
It's perfectly possible to use ```XRegExp(String.raw`^test\raw\slashes$`))``` btw.

treyhunner:
Just chiming in to add that I'm most interested in seeing a special tag in XRegExp so that interpolated strings can be escaped. A `RegExp.escape` proposal was rejected by TC39 last year and it was suggested that a template tag with escaping on interpolation would be more safe than escaping combined with plain string concatenation. Further discussion of this is at benjamingr/RegExp.escape#37.

Until XRegExp supports escaping via a tag, I'll continue my search for a tag function that escapes interpolated strings for use in regular expressions.

josephfrazier:
[...]

PR #180, opened 2017-04-17
Add XRegExp.tag in `build` plugin for tagged template string construction

josephfrazier:
Fixes https://github.com/slevithan/xregexp/issues/103

### DIFF
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
     }
 
     /**
+     * Provides a tag function for building regexes using template literals [1]. See GitHub issue
+     * 103 for discussion [2].
+     *
+     * [1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals
+     * [2]: https://github.com/slevithan/xregexp/issues/103
+     */
+    XRegExp.tag = function (flags) {
+        return function tag (literals /*, ...substitutions */) {
+            var substitutions = [].slice.call(arguments, 1);
+            var subpatterns = substitutions.concat('').map(interpolate);
+            var pattern = literals.raw.map(embedSubpatternAfter).join('');
+            return XRegExp.build(pattern, subpatterns, flags);
+        };
+
+        function interpolate (substitution) {
+            return substitution instanceof RegExp ? substitution : XRegExp.escape(substitution);
+        }
+
+        function embedSubpatternAfter (raw, subpatternIndex) {
+            return raw + '{{' + subpatternIndex + '}}';
+        }
+    };
+
+    /**
      * Builds regexes using named subpatterns, for readability and pattern reuse. Backreferences in
      * the outer pattern and provided subpatterns are automatically renumbered to work correctly.
      * Native flags used by provided subpatterns are ignored in favor of the `flags` argument.
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -61,6 +61,14 @@ module.exports = function(XRegExp) {
             XRegExp(value, flags);
     }
 
+    function interpolate(substitution) {
+        return substitution instanceof RegExp ? substitution : XRegExp.escape(substitution);
+    }
+
+    function embedSubpatternAfter(raw, subpatternIndex) {
+        return raw + '{{' + subpatternIndex + '}}';
+    }
+
     /**
      * Provides a tag function for building regexes using template literals [1]. See GitHub issue
      * 103 for discussion [2].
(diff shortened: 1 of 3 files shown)


src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
     }
 
     /**
+     * Provides a tag function for building regexes using template literals [1]. See GitHub issue
+     * 103 for discussion [2].
+     *
+     * [1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals
+     * [2]: https://github.com/slevithan/xregexp/issues/103
+     */
```
slevithan: Please add an `@example` of using this to the comment

src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
...
+    XRegExp.tag = function (flags) {
```
slevithan: nit: remove space after `function` to match style used by the rest of the library

src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
...
+        return function tag (literals /*, ...substitutions */) {
```
slevithan: nit: remove space after `tag`. probably just remove the function name `tag` as well unless you feel this improves it.

src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
...
+            var substitutions = [].slice.call(arguments, 1);
```
slevithan: Will you be converting this to ES2015 with `...substitutions` and getting rid of this?

src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
...
+            var subpatterns = substitutions.concat('').map(interpolate);
+            var pattern = literals.raw.map(embedSubpatternAfter).join('');
+            return XRegExp.build(pattern, subpatterns, flags);
+        };
+
+        function interpolate (substitution) {
```
slevithan: Please move `interpolate` and `embedSubpatternAfter` out of and above `XRegExp.tag`, to the same level as `asXRegExp`. And remove the spaces after the function names. :)

src/addons/build.js
```diff
@@ -62,6 +62,30 @@ module.exports = function(XRegExp) {
...
+     * Provides a tag function for building regexes using template literals [1]. See GitHub issue
+     * 103 for discussion [2].
+     *
+     * [1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals
+     * [2]: https://github.com/slevithan/xregexp/issues/103
+     */
+    XRegExp.tag = function (flags) {
+        return function tag (literals /*, ...substitutions */) {
+            var substitutions = [].slice.call(arguments, 1);
+            var subpatterns = substitutions.concat('').map(interpolate);
```
slevithan: Feels a little inelegant to always insert an additional unnecessary empty string substitution at the end of `pattern` (enabled here by `concat`-ing an empty string). I'm okay with this as is but if there's a simple way to clean this up, that might be better.

tests/spec/s-addons-build.js
```diff
@@ -1,5 +1,81 @@
 describe('XRegExp.build addon:', function() {
 
+    describe('XRegExp.tag()', function() {
```
slevithan: These tests are great--thanks!

src/addons/build.js
josephfrazier: Yes, I just wanted to make it clear for now. I'll definitely follow up with a PR that uses rest parameters across the codebase.

josephfrazier:
Thanks for the review! I just pushed up a batch of commits that should address your comments. The one about adding an example actually prompted me to add another test and discover/fix a bug. Let me know what you think!

tests/spec/s-addons-build.js
```diff
@@ -74,6 +74,16 @@ describe('XRegExp.build addon:', function() {
...
         });
 
+        it('should work as described in the comment @example', function() {
+            var h12 = /1[0-2]|0?[1-9]/;
+            var h24 = /2[0-3]|[01][0-9]/;
+            var hours = XRegExp.tag('x')`${h12} : | ${h24}`
+            var minutes = /^[0-5][0-9]$/;
+            var time = XRegExp.tag('x')`^ ${hours} (${minutes}) $`
+            expect(time.test('10:59')).toBe(true);
+            expect(XRegExp.exec('10:59', time).minutes).toEqual('59');
```
slevithan: Nit: I'd prefer to use `toBe` wherever strict equality is appropriate, even if just for consistency (`toBe` is used one line up).

src/addons/build.js
```diff
@@ -83,12 +88,12 @@ module.exports = function(XRegExp) {
      * var minutes = /^[0-5][0-9]$/;
      * var time = XRegExp.tag('x')`^ ${hours} (${minutes}) $`
      * time.test('10:59'); // -> true
-     * XRegExp.exec('10:59', time).minutes; // -> '59'
+     * XRegExp.exec('10:59', time)[1]; // -> '59'
```
slevithan: Maybe just remove this line and the capturing group around `minutes` since this is no longer showing an additional feature.

Issue #177, opened 2017-04-17
Add $<n> and $<name> as alternative backreference syntax in replacement text


slevithan:
Add `$<n>` and `$<name>` as alternative backreference syntax for `${n}` and `${name}` in replacement text, so they can be used in ES6 template strings. This would be backward compatible, since `$<` is currently an error in XRegExp replacement text.

See also #103.

PR #181, opened 2017-04-17
Add $<n> and $<name> as alternative backreference syntax in replacement text

josephfrazier:
Fixes https://github.com/slevithan/xregexp/issues/177

I had some trouble using a single regex for the replacement, so I replaced
the $<name> syntax first in a separate pass.

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -51,7 +51,10 @@ var nativeTokens = {
     'class': /\\(?:[0-3][0-7]{0,2}|[4-7][0-7]?|x[\dA-Fa-f]{2}|u(?:[\dA-Fa-f]{4}|{[\dA-Fa-f]+})|c[A-Za-z]|[\s\S])|[\s\S]/
 };
 // Any backreference or dollar-prefixed character in replacement strings
-var replacementToken = /\$(?:{([\w$]+)}|(\d\d?|[\s\S]))/g;
+var replacementTokens = [
+    /\$(?:<([\w$]+)>)/g,
+    /\$(?:{([\w$]+)}|(\d\d?|[\s\S]))/g
+];
 // Check for correct `exec` handling of nonparticipating capturing groups
 var correctExecNpcg = nativ.exec.call(/()??/, '')[1] === undefined;
 // Check for ES6 `flags` prop support
@@ -1590,7 +1593,9 @@ fixed.replace = function(search, replacement) {
         result = nativ.replace.call(this == null ? this : String(this), search, function() {
             // Keep this function's `arguments` available through closure
             var args = arguments;
-            return nativ.replace.call(String(replacement), replacementToken, replacer);
+            return replacementTokens.reduce(function (replacement, replacementToken) {
+                return nativ.replace.call(String(replacement), replacementToken, replacer);
+            }, replacement);
 
             function replacer ($0, $1, $2) {
                 var n;
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -1592,11 +1592,11 @@ fixed.replace = function(search, replacement) {
             var args = arguments;
             return nativ.replace.call(String(replacement), replacementToken, replacer);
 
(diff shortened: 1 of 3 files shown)


src/xregexp.js
```diff
@@ -51,7 +51,10 @@ var nativeTokens = {
     'class': /\\(?:[0-3][0-7]{0,2}|[4-7][0-7]?|x[\dA-Fa-f]{2}|u(?:[\dA-Fa-f]{4}|{[\dA-Fa-f]+})|c[A-Za-z]|[\s\S])|[\s\S]/
 };
 // Any backreference or dollar-prefixed character in replacement strings
-var replacementToken = /\$(?:{([\w$]+)}|(\d\d?|[\s\S]))/g;
+var replacementTokens = [
+    /\$(?:<([\w$]+)>)/g,
+    /\$(?:{([\w$]+)}|(\d\d?|[\s\S]))/g
+];
```
slevithan: I'd rather do this in one pass (rather than 2 or 3) for speed. But that aside, as written it doesn't make sense to do this in two passes since you're doing the same handling for `$2` in the `replacer` function even when it isn't relevant.

I think what you want here is to make `replacementToken` `/\$(?:{([\w$]+)}|<([\w$]+)>|(\d\d?|[\s\S]))/g`, then change `replacer ($0, $1, $2)` to `replacer($0, $1, $2, $3)` or something like `replacer($0, bracketed, angled, dollarToken)`, then put `bracketed = bracketed || angled` at the top of the `replacer` function.

josephfrazier: Ah, that's exactly right! Thanks for the suggestion, I've pushed some updates implementing it.

src/xregexp.js
```diff
@@ -1594,11 +1594,11 @@ fixed.replace = function(search, replacement) {
             var args = arguments;
             return nativ.replace.call(String(replacement), replacementToken, replacer);
 
-            function replacer ($0, $1, $2, $3) {
-                $1 = $1 || $2
+            function replacer ($0, bracketed, angled, dollarToken) {
+                bracketed = bracketed || angled
```
slevithan: <nit> Please add a trailing semicolon

src/xregexp.js
```diff
@@ -1157,6 +1157,8 @@ XRegExp.matchChain = function(str, chain) {
  *       backreference n/nn.
  *     - ${n} - Where n is a name or any number of digits that reference an existent capturing
  *       group, inserts backreference n.
+ *     - $<n> - Where n is a name or any number of digits that reference an existent capturing
```
slevithan: See above where `$&, $0` are on the same line. Would you mind doing the same here, and grouping this with `${n}`?

tests/spec/s-xregexp-natives.js
```diff
@@ -508,13 +508,16 @@ describe('When overridden, String.prototype.replace()', function() {
 
                 it('should return the named backreference', function() {
                     expect('test'.replace(XRegExp('(?<test>t)', 'g'), ':${test}:')).toBe(':t:es:t:');
+                    expect('test'.replace(XRegExp('(?<test>t)', 'g'), ':$<test>:')).toBe(':t:es:t:');
 
                     // Backreference to a nonparticipating capturing group
                     expect('test'.replace(XRegExp('t|(?<test>t)', 'g'), ':${test}:')).toBe('::es::');
+                    expect('test'.replace(XRegExp('t|(?<test>t)', 'g'), ':$<test>:')).toBe('::es::');
                 });
```
slevithan: Can you add a test for mixing `${n}` and `$<n>` syntax in the same replacement string?

slevithan:
This is looking good.

In addition to the few small comments I just added, can you rebase this and fix the conflict?

> This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

PR #183, opened 2017-04-25
Revert isPatternNext to isQuantifierNext

josephfrazier:
This makes the "Constructor with x flag, whitespace, and comments" test
fast again. From
https://github.com/slevithan/xregexp/pull/164#issuecomment-294345042:

> A couple ideas: avoid the string concatenation in `isPatternNext`
> (possibly going back to regex literals and making the function specific
> to quantifiers again even though the current code is more
> readable/maintainable, since this isn't needed to handle simple cases
> with whitespace followed by `)`)

Since [babel-plugin-transform-xregexp] automatically compiles the `new
RegExp()` calls into literals, we get (most of) the performance back
without sacrificing the readability of having separate subpatterns.

[babel-plugin-transform-xregexp]: https://www.npmjs.com/package/babel-plugin-transform-xregexp

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,5 @@
 {
+  "plugins": ["transform-xregexp"],
   "presets": [["env", {
     "exclude": ["transform-es2015-literals"]
   }]]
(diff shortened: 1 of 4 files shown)


slevithan:
Awesome

PR #187, opened 2017-04-26
.gitignore compiled files, build them on prepublish

josephfrazier:
Following up on https://github.com/slevithan/xregexp/pull/182#issuecomment-297198961,
this change removes `lib/`, `xregexp-all.js`, and `bower.json` from the
repository.

It also adds a `prepublish` [script] that builds/tests the files before
publishing to npm, to ensure they are included. You can run [npm pack]
to see the `prepublish` script in action, then you can take a look at
the generated tarball with `tar tvf xregexp-3.2.0.tgz` and verify that
`lib/` and `xregexp-all.js` are still there.

[script]: https://docs.npmjs.com/misc/scripts
[npm pack]: https://docs.npmjs.com/cli/pack


kaidjohnson:
To be clear, this retains `xregexp-all.js` in the npm repo but *not* in the git repo, correct?

josephfrazier:
Yes, both `xregexp-all.js` and the entire `lib/` directory will be present in the packages published to npm, but they will not be tracked in the git repo.

kaidjohnson:
Does this affect bower at all? I realize bower is 'on the way out', but I'm sure many application maintainers are still using it in their daily build flows.

josephfrazier:
Yes, as suggested in the [comment](https://github.com/slevithan/xregexp/pull/182#issuecomment-297198961) this PR addresses, future releases will no longer be available via bower. This is because [bower uses files from the git repo](https://bower.io/docs/creating-packages/#register), and this removes both `bower.json` and the file it points to, `xregexp-all.js`, from the git repo.

However, we might be able to restore bower availability by adding bower.json back, setting its [main field] to `src/index.js`, and setting the [moduleType field] field to either `node` or `es6` (if we upgrade to ES6 modules). Note that bower actually recommends using unminified source files over bundles, in the [main field] documentation:

> * Use source files with module exports and imports over pre-built distribution files.
> * Do not include minified files.

[main field]: https://github.com/bower/spec/blob/59c8f0e8f8444cbdd71a091919d28d0fa29d56fe/json.md#main
[moduleType field]: https://github.com/bower/spec/blob/59c8f0e8f8444cbdd71a091919d28d0fa29d56fe/json.md#moduletype

Issue #108, opened 2015-09-27
Stop supporting pre-ES5 browsers


slevithan:
Pre-ES5 browsers include e.g. IE 8 and earlier. Right now XRegExp still supports ancient browsers like IE 5.5, Firefox 1.5, Chrome 1, and Safari 3, which is totally unnecessary. But there's not much benefit to dropping support for these browsers until XRegExp can rely on ES5.

Some things to do after dropping pre-ES5 support:
- Generally take advantage of any ES5 features.
  - Use character indexing on strings instead of `charAt`, use `Array.prototype.indexOf`, etc.
- Possibly kill the entire custom split implementation and make `XRegExp.split` just defer to `String.prototype.split`, if there are few enough edge cases that ES5 browsers get wrong.
- Possibly kill handling for inconsistencies related to nonparticipating capturing groups (search for `nonparticipating` in code), if ES5 browsers generally get this right.
- Possibly kill the `[^]` and `[]` transforms if all ES5 browsers get this right.

josephfrazier:
On a related note, would you be interested in using [babelify](https://github.com/babel/babelify) to compile ES2015+ code down to ES5? This would allow a [variety of syntax upgrades](https://babeljs.io/learn-es2015/) to the files in `src/`, while preserving correct behavior of `xregexp-all.js` in all ES5+ browsers.

One way of doing this is to install `babelify` and `babel-preset-es2015` as `devDependencies`, then change the `build` script to

```sh
browserify src/index.js --standalone XRegExp --transform [ babelify --presets [ es2015 ] ] > xregexp-all.js
```

slevithan:
@josephfrazier Yes I'd be happy happy to move to ES2015+ code and compile down to ES5, if you're willing to set it up! :D Before doing that I'd want to get out a minor update, since dropping support for pre-ES5 browsers should be a v4.0.0 breaking change.

It would be nice to also have ES5 versions of all the individual modules, for people who don't want to use `xregexp-all.js`. I guess those could go in a new `dist` directory or similar, and `xregexp-all.js` should probably move in there as well.

josephfrazier:
[...]

PR #194, opened 2017-04-30
Convert string concatenation to template strings

josephfrazier:
As mentioned in https://github.com/slevithan/xregexp/issues/108#issuecomment-280100000

This was done with the help of [lebab](https://github.com/lebab/lebab)
(with the patch at https://github.com/lebab/lebab/pull/243),
using `lebab --replace src --transform template`.

### DIFF
--- a/src/addons/build.js
+++ b/src/addons/build.js
@@ -187,7 +187,7 @@ export default (XRegExp) => {
                     intro = '(?:';
                 }
                 numPriorCaps = numCaps;
-                return `${intro + data[subName].pattern.replace(subParts, (match, paren, backref) => {
+                const rewrittenSubpattern = data[subName].pattern.replace(subParts, (match, paren, backref) => {
                     // Capturing group
                     if (paren) {
                         capName = data[subName].names[numCaps - numPriorCaps];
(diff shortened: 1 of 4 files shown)


src/addons/build.js
```diff
@@ -174,56 +174,56 @@ export default (XRegExp) => {
...
                     // If it's a named group, preserve the name. Otherwise, use the subpattern name
                     // as the capture name
-                    intro = '(?<' + (capName || subName) + '>';
+                    intro = `(?<${capName || subName}>`;
                 } else {
                     intro = '(?:';
                 }
                 numPriorCaps = numCaps;
-                return intro + data[subName].pattern.replace(subParts, (match, paren, backref) => {
+                return `${intro + data[subName].pattern.replace(subParts, (match, paren, backref) => {
```
slevithan: This feels really excessive. At a minimum, this should separately interpolate `intro` and the rest of this. But really, this should probably not use a `` ` `` string at all, unless the rest of this is first assigned to a variable.

josephfrazier: Yeah, I was a bit iffy on that one too... I just pushed https://github.com/slevithan/xregexp/pull/194/commits/b1def280d73a8a6ec96b6b69d2499426c0525037 implementing the suggestion of an extra variable. Let me know what you think.

PR #195, opened 2017-05-01
Break large regexes into multiline template strings

josephfrazier:
[babel-plugin-transform-xregexp] v0.0.6 uses XRegExp to compile
`RegExp()` calls with a template literal argument into the corresponding
regex literal. This compile-time transformation approximates the
functionality of `XRegExp.tag`, and doesn't have runtime overhead. See
here for more details:
https://github.com/josephfrazier/babel-plugin-transform-xregexp/commit/2a0100003c1398d4

When combined with the 'x' flag support that XRegExp offers, we can use
this to break long regex literals across multiple lines, without
introducing any extra escaping. If you run `git show --color-words=.` on
the first commit in this PR, you'll see that the patterns don't change
at all, other than having whitespace inserted for readability.

The remaining commits simply do a little deduplication in patterns.

[babel-plugin-transform-xregexp]: https://www.npmjs.com/package/babel-plugin-transform-xregexp


slevithan:
The thinking behind this is cool, but implicitly changing the syntax of `new RegExp` is difficult-to-understand magic, and should *at least* be accompanied by comments explaining what’s going on. It might not be worthwhile at all given that contributors to XRegExp are generally able to read regexes more comfortably than most.

(XRegExp behavior already has some of its own unintuitive magic, but I'd like to trend toward removing nonobvious behavior/features.)

slevithan:
Even in the updated form, I think it might not be worth doing this in the core XRegExp library. :'( If there are places we can use `XRegExp.tag` directly, I'd be in favor of that.

josephfrazier:
Yeah... it's still a bit magical, since the tagged template literal processing wouldn't allow matching a literal space character, were that to be desired. Using `XRegExp.tag` directly is tricky, because the only file that could do it without having a recursive `import` would be `src/addons/unicode-base.js`, but even then it would imply that anyone who wants to use that plugin also has to have the `build` plugin included.

I might try one more approach, just to see how it looks, but won't mind at all if we don't land it.

Issue #179, opened 2017-04-17
Reduce cases where unnecessary (?:) separator is added to pattern


slevithan:
Update the XRegExp constructor (more specifically the `runTokens` function) to pass preceding tokens to token handler functions. Alternatively they could be made available on the token handler context (like the existing `this.captureNames` and `this.hasNamedCapture`).

In addition to this being a generally useful feature that could enable more powerful syntax addons, this would enable much better handling of where `(?:)` separators are inserted when stripping whitespace (with flag `x`) and comments out of a pattern. Currently the `getContextualTokenSeparator` function is used for this, but it's not very robust. E.g. it avoids adding `(?:)` if the preceding character is `(`, but it doesn't deal with `(?<name>` and other cases where a separator isn't needed.

It would make things even easier if preceding tokens were annotated with a `type` string (provided as a new property on the `options` argument of `XRegExp.addToken`). E.g. from `XRegExp('(?<name>\n.)', 'x')` you could get:

```js
[
  {type: 'named-capture-start', output: '('},
  {type: 'x-ignored', output: ''},
  {type: 'native-token', output: '.'},
  {type: 'native-token', output: ')'}
]
```
Then the `getContextualTokenSeparator` function can easily check whether the preceding token is something that requires the `(?:)` separator be added.

This above idea of making preceding tokens available to token handler functions would require special handling for the `reparse` option, since only the final reparsed version of token should be added to the list of prior tokens.

After this change:
* Check whether the lines in `build.js` for handling `(?:)` are still needed.
* Consider removing the cleanup of double `(?:)(?:)`.
* Upgrade to support at least `(?<name>` as an additional case when `(?:)` doesn't need to be added.

See related discussion in <https://github.com/slevithan/xregexp/pull/164#issuecomment-294345042>.

slevithan:
This sort of structure could also be the foundation of a useful advanced feature from PCRE, Perl, etc.: The ability to reference the entire *contents* of a named group (including nested parens) from later in the pattern, enabling support for subpattern reuse via `(?&name)` and `(?n)` (aka subpatterns as subroutines).

This would simply require generic syntax tokens for `)` and any `(` that isn't part of a self-contained token like `(?#...)` to mark subsequent tokens as children until the closing `)` arrives. Then the generated pattern contents of each named group could be derived when needed.

Perhaps this would look like:

```js
[
  {
    type: 'named-capture-start',
    output: '(',
    children: [
      {type: 'x-ignored', output: ''},
[...]

PR #196, opened 2017-05-01
Cleanup more `(?:)` from patterns

josephfrazier:
Following up on https://github.com/slevithan/xregexp/pull/164, this
change prevents a `(?:)` from being inserted in the following places:

* At the beginning of a non-capturing group or (negated) lookahead (the end is already handled)
* Before or after a `|`
* At the beginning or the end of the pattern

This solution isn't as complete as the one suggested in
https://github.com/slevithan/xregexp/issues/179, but it's a decent
stopgap.

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -238,8 +238,8 @@ function getContextualTokenSeparator(match, scope, flags) {
         // No need to separate tokens if at the beginning or end of a group
         match.input[match.index - 1] === '(' ||
         match.input[match.index + match[0].length] === ')' ||
-        // No need to separate tokens if at the beginning of a non-capturing group
-        match.input.slice(match.index - 3, 3) === '(?:' ||
+        // No need to separate tokens if at the beginning of a non-capturing group or lookahead
+        match.input.slice(match.index - 3, 3).match(/\(\?[:=!]/) ||
         // No need to separate tokens if before or after a `|`
         match.input[match.index - 1] === '|' ||
         match.input[match.index + match[0].length] === '|' ||
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -247,8 +247,8 @@ function getContextualTokenSeparator(match, scope, flags) {
         match.input[match.index - 1] === '|' ||
         match.input[match.index + match[0].length] === '|' ||
         // No need to separate tokens if at the beginning or end of the pattern
-        match.input[match.index - 1] === undefined ||
-        match.input[match.index + match[0].length] === undefined ||
+        match.index - 1 < 0 ||
+        match.index + match[0].length >= match.input.length ||
         // Avoid separating tokens when the following token is a quantifier
         isQuantifierNext(match.input, match.index + match[0].length, flags)
     ) {
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -238,6 +238,14 @@ function getContextualTokenSeparator(match, scope, flags) {
         // No need to separate tokens if at the beginning or end of a group
         match.input[match.index - 1] === '(' ||
         match.input[match.index + match[0].length] === ')' ||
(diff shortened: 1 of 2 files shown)


josephfrazier:
Any thoughts on this one, @slevithan? I'd like to land it if there's no objections.

src/xregexp.js
```diff
@@ -238,8 +238,8 @@ function getContextualTokenSeparator(match, scope, flags) {
         // No need to separate tokens if at the beginning or end of a group
         match.input[match.index - 1] === '(' ||
         match.input[match.index + match[0].length] === ')' ||
-        // No need to separate tokens if at the beginning of a non-capturing group
-        match.input.slice(match.index - 3, 3) === '(?:' ||
+        // No need to separate tokens if at the beginning of a non-capturing group or lookahead
+        match.input.slice(match.index - 3, 3).match(/\(\?[:=!]/) ||
```
slevithan: What happens if `match.index - 3` is e.g. `-1`? Even if this works, it's not precise in its meaning so might be better to clarify.

Also, this should be using `nativ.match` or preferably `nativ.test` (`nativ.test.call(regex, str)`), at least as long as those still exist. XRegExp never uses native regex methods directly since (until v4.0.0) the native methods could be overridden. As a minor optimization, the regex should start with `^` so it isn't tested at all of the string's character positions if it fails to match.

src/xregexp.js
```diff
@@ -238,6 +238,14 @@ function getContextualTokenSeparator(match, scope, flags) {
         // No need to separate tokens if at the beginning or end of a group
         match.input[match.index - 1] === '(' ||
         match.input[match.index + match[0].length] === ')' ||
+        // No need to separate tokens if at the beginning of a non-capturing group
+        match.input.slice(match.index - 3, 3) === '(?:' ||
+        // No need to separate tokens if before or after a `|`
+        match.input[match.index - 1] === '|' ||
+        match.input[match.index + match[0].length] === '|' ||
+        // No need to separate tokens if at the beginning or end of the pattern
+        match.input[match.index - 1] === undefined ||
```
slevithan: Can you just check the value of `match.index - 1` directly, without involving `match.input`?

src/xregexp.js
```diff
@@ -238,6 +238,14 @@ function getContextualTokenSeparator(match, scope, flags) {
...
+        match.input[match.index + match[0].length] === undefined ||
```
slevithan: Ditto. Should probably be comparing to the length of `match.input` rather than referencing a specific character in the string.

src/xregexp.js
josephfrazier: Good point, fixed in https://github.com/slevithan/xregexp/pull/196/commits/ee766c211250cc4c90d73123ff1d6095007f0c96 and https://github.com/slevithan/xregexp/pull/196/commits/222eda9017dc424a18b9ff3c26cb04d71946f30d

src/xregexp.js
josephfrazier: Also fixed in https://github.com/slevithan/xregexp/commit/ee766c211250cc4c90d73123ff1d6095007f0c96

src/xregexp.js
josephfrazier: Thanks for pointing this out, I switched to `nativ.test` in https://github.com/slevithan/xregexp/pull/196/commits/2bdd98250a7cc08e60e56a59c0b362f6657ce1ee. Also, it turns out that `slice` was the wrong method to use, so I switched to `substr` in https://github.com/slevithan/xregexp/pull/196/commits/75c8402d30692f77491b7222e4cb58b607673029, and documented its behavior with negative indices in https://github.com/slevithan/xregexp/pull/196/commits/fd9940b84888490e1b85a33afc31366bb68a69b2.

Issue #89, opened 2015-06-28
Remove installable 'natives' feature that modifies String and RegExp prototypes


retorquere:
I'm including xregexp in a firefox extension, and the changes to RegExp.prototype and String.prototype don't pass the impending extension signature validation. Is it possible to get an xregexp build that doesn't make these modifications to these prototypes?

slevithan:
Hm, interesting. There's very little benefit to adding this option, since it would save almost no code, the "natives" option is already self-contained and not relied on by the rest of the code, and native overrides are already off by default. You'd essentially just need to comment out https://github.com/slevithan/xregexp/blob/b867633a27b0418c5ad8ae2e4f7619d89cc6bdb0/src/xregexp.js#L376-380 or kill the `setNatives` function and its calls in `install` and `uninstall`. I'd recommend doing that for now, but will leave this open for more feedback.

retorquere:
Fair enough. I'm not saying the addons validator is sane, but as it's going
to be enforced on all Firefox extensions, I have to dance to its tune.

slevithan:
Would be happy to see a pull request that showed how this would work.

retorquere:
That's in essence what my pull request would do.

wilmoore:
> I'm also considering removing support for XRegExp.install('natives') in the next major breaking change (XRegExp 4.0.0), which would resolve this since it would remove the lines in question for everyone.

Before reading the above I was going to suggest something similar. Move `XRegExp.install('natives')` into a separate package that can optionally be installed by anyone who cared to.

slevithan:
`XRegExp.install('natives')` was marked as deprecated in commit 57b3172cefda646b8ae92027a4cf83b55daae823.

retorquere:
Super. Any timeline on a release which would contain this change?

slevithan:
This will be included in v4.0.0. Today, v3.1.0 was released. Sorry, no timeline on v4 at the moment.

There's no required delay before releasing v4.0.0, but I would like to bundle the drop of support for `XRegExp.install('natives')` with at least one other planned breaking change: #108 (drop all support for pre-ES5 browsers). Others might end up included as well. These changes might take a while since I'm not able to dedicate much time to the project at the moment.

slevithan:
After making this change, will also want to update `tests/helpers/h.js` to no longer try to uninstall `'natives'`.

slevithan:
@retorquere this is finally completed and included in v4.0.0, thanks to @josephfrazier.

PR #207, opened 2018-01-01
Remove support for `XRegExp.install('natives')`

josephfrazier:
This fixes https://github.com/slevithan/xregexp/issues/89


tests/perf/perf.js
```diff
@@ -110,9 +110,7 @@
         var str = Array(30 + 1).join('hello world x ') + 'xx!';
         var pos = 5;
 
-        XRegExp.install('natives');
-        var fixedExec = RegExp.prototype.exec;
-        XRegExp.uninstall('natives');
+        var fixedExec = XRegExp.exec;
```
slevithan: The 3 cases below of comparing "shimmed exec" to XRegExp.exec are no longer testing different things. I'd recommend removing the `fixedExec` var entirely and the 3 `.add` calls that use it.

josephfrazier: Oh, good point! I was essentially doing a naive search/replace on the code, so I didn't notice this. Fixed in https://github.com/slevithan/xregexp/pull/207/commits/9f21bd0c5f0119c3b23b99bc7210061e49887bee

Issue #175, opened 2017-04-17
Move named backreferences to a groups object


slevithan:
Right now, looping over named backreferences after executing a regex with named groups is unnecessarily difficult, since XRegExp augments match arrays with the named backreferences as direct properties. It would be nice to give access to a clean object with named groups.

Proposal:

* Add a new option "namespacing" to `XRegExp.install`/`uninstall`/`isInstalled`.
* This option would be off by default in the initial version when it is introduced. In the subsequent major version (to enable a breaking change), the "namespacing" option will be on by default, and the ability to turn it off might be removed.
* When enabled, named backreferences are no longer added as direct properties of match arrays. Rather, they are added to a `groups` object on result arrays.
  * For `String.prototype.replace` replacement functions (where named backreferences are added as properties of the first String argument), maybe leave things as they are (no `groups` object). If https://github.com/tc39/proposal-regexp-named-groups moves beyond the proposal stage, then XRegExp should follow wherever ECMAScript is headed.
* When enabled, `length` and `__proto__` should no longer be reserved words for named captures. Using `__proto__` as a capture name would still lead to undesired results, so don’t do that.

---
As a workaround for now, here's how you can get a clean object with named groups:

```js
function groups(regex, match) {
    var o = {};
    var x = regex.xregexp;
    (x && x.captureNames || []).forEach(function(name) {
        if (name) {
            o[name] = match[name];
        }
    });
    return o;
}

// Use it
var regex = XRegExp('(?s)(?<char>.)');
var match = XRegExp.exec('test', regex);
groups(regex, match);
// -> {char: 't'}
```

---
Moved here from https://github.com/slevithan/xregexp/issues/45#issuecomment-292875730 since that was a long thread.

dploeger:
I like, as this would make the type definitions for things like Typescript much better. I currently have to use an "any-cast", which is very bad practice in Typescript:

```typescript
let pattern = XRegExp('a(?<groupName>.)c')
let match = XRegExp.exec('abc', pattern)
console.log((match as any).groupName)
```

(because "groupName" is not a defined property in the "RegExpExecArray" type, used by the type definitions for XRegExp over at definitelyTyped) and extending it with dynamic properties is nearly impossible or inducing any-types, again)

josephfrazier:
> * When enabled, `length` and `__proto__` should no longer be reserved words for named captures. Using `__proto__` as a capture name would still lead to undesired results, so don’t do that.

[...]

PR #210, opened 2018-01-15
[WIP] Add `namespacing` feature for named capture groups object

josephfrazier:
This addresses https://github.com/slevithan/xregexp/issues/175

I still need to add proper tests, but the new behavior can be
demonstrated by running this after `npm run build`

```js
var XRegExp = require('./')
XRegExp.install('namespacing')
var regex = XRegExp('(?s)(?<__proto__>.)');
var match = XRegExp.exec('test', regex);
console.log(match)

const replaced = XRegExp.replace('test', regex, function (matched, capture1, position, S, groups) {
    console.log({groups})
})
```

### DIFF
--- a/src/xregexp.js
+++ b/src/xregexp.js
@@ -1403,12 +1403,17 @@ fixed.exec = function(str) {
         }
 
         // Attach named capture properties
+        let namedCaptureObject = match;
+        if (XRegExp.isInstalled('namespacing')) {
+            match.groups = Object.create(null); // https://tc39.github.io/proposal-regexp-named-groups/#sec-regexpbuiltinexec
+            namedCaptureObject = match.groups;
+        }
         if (this[REGEX_DATA] && this[REGEX_DATA].captureNames) {
             // Skip index 0
             for (let i = 1; i < match.length; ++i) {
                 const name = this[REGEX_DATA].captureNames[i - 1];
                 if (name) {
-                    match[name] = match[i];
+                    namedCaptureObject[name] = match[i];
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/xregexp.js
```diff
@@ -916,18 +928,25 @@ XRegExp.globalize = (regex) => copyRegex(regex, {addG: true});
  * // With an options object
  * XRegExp.install({
  *   // Enables support for astral code points in Unicode addons (implicitly sets flag A)
- *   astral: true
+ *   astral: true,
+ *
+ *   // Enables named capture groups
```
slevithan: This seems misleading, since named capture groups are always enabled. What about "``Moves named backreferences to the `groups` property of match arrays``", "``Use a `groups` object to store named backreferences``", or something similar?

src/xregexp.js
```diff
@@ -1250,18 +1270,25 @@ XRegExp.test = (str, regex, pos, sticky) => !!XRegExp.exec(str, regex, pos, stic
  * // With an options object
  * XRegExp.uninstall({
  *   // Disables support for astral code points in Unicode addons
- *   astral: true
+ *   astral: true,
+ *
+ *   // Disables named capture groups
```
slevithan: Ditto.

src/xregexp.js
josephfrazier: Good catch! In https://github.com/slevithan/xregexp/pull/210/commits/24a29d0b2e4a7fa5aad0da028cce33521fe9863a ,I decided to go with:

    Adds named capture groups to the `groups` property of match arrays.

since "capture groups" seemed a little more accurate than "backreferences", and it matches the wording in the TC39 proposal.

josephfrazier:
Great, I just added some tests, let me know what you think!

PR #211, opened 2018-01-20
Run eslint on tests

josephfrazier:
This helps the tests conform to the same style as the library,
with the exception of a handful of rules that are not enforced.


slevithan:
Nice, thanks!

There's a conflict in `h.js` after merging the `namespacing` PR. Can you resolve?

Issue #215, opened 2018-01-23
Upgrade to Unicode 10


slevithan:
The data for XRegExp's Unicode plugins currently uses Unicode 9.0.0.

slevithan:
See #171 for upgrade steps.

josephfrazier:
https://github.com/slevithan/xregexp/issues/171#issuecomment-293201412:

> It might be worth considering rewriting the build scripts to consume the node-unicode-data packages, e.g. https://github.com/mathiasbynens/unicode-9.0.0, directly. This would greatly simplify things and could reduce the manual work required for such updates in the future.

I might be interested in taking this approach, if it's ok with you. I'd probably start by rewriting the existing script to use `unicode-9.0.0` and make sure that the tests pass and that the built files are equivalent, then I would upgrade to `unicode-10.0.0`.

slevithan:
That sounds fantastic!

mathiasbynens:
@josephfrazier Yes please! Go for it, and feel free to ping me if you run into any issues.

Issue #171, opened 2017-04-11
Upgrade to Unicode 9.0.0


slevithan:
From <http://www.unicode.org/versions/Unicode9.0.0/>:

> Unicode 9.0 adds exactly 7,500 characters, for a total of 128,172 characters. These additions include six new scripts and 72 new emoji characters.

The categories, scripts, blocks, and properties addons will all need to be updated. Steps:

* Delete files in `tools/data/` for the old version (8.0.0).
* Update the Unicode version number in `tools/generate-unicode.sh`.
* Manually update the code points and ranges for the handful of properties in `tools/scripts/property-regex.py`, using data generatable from [UnicodeSet](http://unicode.org/cldr/utility/list-unicodeset.jsp) (see [comments here](https://github.com/slevithan/xregexp/issues/29#issuecomment-6534190)).
* Run `tools/generate-unicode.sh`.
* Use the generated data in `tools/output/` to update the Unicode addon JS files in `src/addons/`.

Also update the "Uses Unicode 8.0.0" comments in `README.md`, `unicode-blocks.js`, `unicode-categories.js`, `unicode-properties.js`, and `unicode-scripts.js`.

mathiasbynens:
It might be worth considering rewriting the build scripts to consume the node-unicode-data packages, e.g. https://github.com/mathiasbynens/unicode-9.0.0, directly. This would greatly simplify things and could reduce the manual work required for such updates in the future.

slevithan:
If nothing else, it might be easier to get the data for the handful of properties listed in `tools/scripts/property-regex.py` from https://github.com/mathiasbynens/unicode-9.0.0/tree/master/Binary_Property than from UnicodeSet.

PR #219, opened 2018-02-12
Import data from tools/output/*js instead of copy/pasting

josephfrazier:
This makes the upgrade process a little easier, by removing the following step from https://github.com/slevithan/xregexp/issues/171#issue-220875142:

> * Use the generated data in `tools/output/` to update the Unicode addon JS files in `src/addons/`.

It will also make https://github.com/slevithan/xregexp/issues/215#issuecomment-364590883 easier,
as the `tools/output/*js` files can be rewritten to use `unicode-9.0.0`
directly (if it's space-efficient).


mathiasbynens:
> It will also make #215 (comment) easier, as the tools/output/*js files can be rewritten to use unicode-9.0.0 directly (if it's space-efficient).

The `unicode-*` packages are not space-efficient, but they don’t have to be. What I had in mind is that the `tools/output/*.js` files would be *generated* by a simple `*.js` build script that uses the `unicode-*` data directly. We could then remove all the Python scripts from this repo.

PR #184, opened 2017-04-25
Add eslint to enforce code style

josephfrazier:
This will make it easier to automatically enforce the pre-existing code
style of the project, using [eslint]. It was done by running:

    npm install --save-dev eslint
    ./node_modules/.bin/eslint --init

[eslint]: https://github.com/eslint/eslint/

This script interactively generates an eslint configuration (see [here]).
Here's how I answered its questions:

* Choosing `Inspect your JavaScript file(s)`
* Entering `src` as the files to inspect
* Choosing a JavaScript config file format
* Answering yes to ES6 features/classes
* Choosing both Node/Browser as environment
* Answering yes to CommonJS
* Answering no to JSX/React

[here]: http://devnull.guru/adding-eslint-to-your-project-is-easier-than-ever/

After that, I added the `lint` script to package.json, and manually
changed the `space-before-function-paren` to:

    "space-before-function-paren": [
        "error",
        "never"
    ],

to enforce the rule mentioned here:

* https://github.com/slevithan/xregexp/pull/180#discussion_r113109471
* https://github.com/slevithan/xregexp/pull/180#discussion_r113109489

PR #220, opened 2018-02-14
Simplify Unicode build scripts

mathiasbynens:
This uses node-unicode-data instead of my old Python scripts, reducing the need to maintain the Unicode data parser as part of this repository.


package.json
```diff
@@ -42,6 +43,8 @@
     "browserify": "^12.0.1",
     "eslint": "^3.19.0",
     "jasmine": "^2.5.3",
+    "jsesc": "^2.5.1",
+    "unicode-9.0.0": "^0.7.4",
```
mathiasbynens: I’ve used unicode-9.0.0 (and not unicode-10.0.0) here to more clearly show the actual diff. We can easily update to unicode-10.0.0 later by just swapping out the dependency (i.e. changing this line) and then running `npm run build-unicode-data` again. No more manually updating Unicode data files and Python scripts!

tools/output/blocks.js
```diff
@@ -1,1094 +1,1094 @@
 module.exports = [
-        {
-            name: 'InAdlam',
-            astral: '\uD83A[\uDD00-\uDD5F]'
```
mathiasbynens: The removal of `[` and `]` is a bug. Looking into it…

tools/output/blocks.js
```diff
@@ -1,1094 +1,1094 @@
...
-        {
-            name: 'InCJK_Radicals_Supplement',
-            bmp: '\u2E80-\u2EFF'
-        },
-        {
-            name: 'InCJK_Strokes',
-            bmp: '\u31C0-\u31EF'
-        },
-        {
-            name: 'InCJK_Symbols_and_Punctuation',
```
mathiasbynens: Turns out the canonical name is `InCJK_Symbols_And_Punctuation`, with a capital letter `A`. This is technically a breaking change.

tools/output/blocks.js
mathiasbynens: Ah, forgot to flip a flag after debugging. Fixed.

josephfrazier:
Thanks for working on this! I had been thinking about it, but I hadn't yet worked out how to go about the port. For example, using `jsesc` and `unicode-property-value-aliases` would have taken me a while to figure out. I'll comment further inline.

tools/output/blocks.js
josephfrazier: Can we make it preserve the pre-existing names, so as to avoid the breaking change for now? Of course, we should probably open a separate issue about using the canonical names.

package.json
josephfrazier: Yeah, I think this is the way to go! I may also look into using a modified version of [prettier-diff](https://github.com/josephfrazier/prettier-diff) to ignore differences in the object key ordering.

mathiasbynens: I tried to make sure the order remains the same. AFAICT the only difference is that the property names are now quoted (jsesc does this), i.e. `name: 'Foo'` becomes `'name': 'Foo'`, and that the escaping for various characters is different.

tools/scripts/utils.js
```diff
@@ -132,7 +132,7 @@ const createBmpRange = (r, { addBrackets } = { addBrackets: true }) => {
 
 const assemble = ({ name, alias, codePoints }) => {
     const { bmp, astral, isBmpLast } = createRange(codePoints);
-    const result = { name };
+    const result = { name: name.replace('_And_', '_and_').replace('_For_', '_for_') };
```
mathiasbynens: FWIW, I’m against this change. IMHO we should use [the canonical names](https://unicode.org/Public/10.0.0/ucd/PropertyValueAliases.txt), not an arbitrarily-cased variation of them. I understand the desire to merge this with the workaround but I’m hoping it can be removed later.

josephfrazier: Absolutely, we should ultimately use the canonical names :) I just want to keep the refactoring separate from behavioral changes.

tools/output/blocks.js
slevithan: It's not a breaking change to change casing. Feel free to update it to the canonical name! Per Unicode's recommendations, XRegExp already treats Unicode property names as case insensitive, with any spaces, hyphens, and underscores ignored. See https://github.com/slevithan/xregexp/blob/263ad8ca9963d4b81d10583f6e1edb446a88ad26/src/addons/unicode-base.js#L35-L38

mathiasbynens:
@josephfrazier Thanks for the help here! I’ve squashed all commits to remove the casing changes that are not needed per @slevithan’s latest comment.

tools/output/blocks.js
mathiasbynens: Oh, right! Woohoo :) (I was thinking about native Unicode property escapes, where I explicitly decided *not* to normalize.)

Patch updated.

josephfrazier: Ooh, I wasn't aware of that. Nice!

josephfrazier:
Thanks @mathiasbynens ! Now that we're using JS for the build scripts, I went ahead and added linting. We can always squash upon merge, instead of having to repeatedly force-push the branch. Let me know what you think.

mathiasbynens:
I disagree with the linter on some stylistic choices 😅 But your changes LGTM. Thanks!

josephfrazier:
FWIW, the lint rules were only set up to enforce the pre-existing project style, and it sounds like @slevithan  is open to changing the style (see https://github.com/slevithan/xregexp/pull/184), but we can do that separately.

Thanks for the review! I think I'm comfortable with the changes here, but would like @slevithan to sign off as well. If it helps, I found that using [prettier-diff](https://github.com/josephfrazier/prettier-diff) makes it easier to ignore the stylistic changes to `tools/output/*js`. The semantic changes appear to be:

* Adding `Cased_Letter` to `categories.js`
* Adding a lot of values to `properties.js`
* Unescaping some characters in `bmp` values throughout

slevithan:
This looks awesome!

`Cased_Letter` is nice to add.

I'm not confident about adding all of the binary properties in `unicode-properties.js`. Many of them do not seem very useful (of course this is subjective), and a few of those that seem more useful already have similar equivalents or can be constructed using `unicode-categories.js`. The list so far has intentionally been only those 9 binary properties needed to meet the UTS 18 Level 1 RL1.2 requirements for Unicode regex support. See <http://unicode.org/reports/tr18/#RL1.2>, and note that `Assigned` is already supported through the special handling at <https://github.com/slevithan/xregexp/blob/d2e73199fd1bb087a0c7e9b0a19ec03ffe56e63c/src/addons/unicode-properties.js#L58-L63>.

But I'm fine with deferring to @mathiasbynens on this. Mathias, if you give the confirmation here that you think it's the right call to add them all, then we can move forward with that. I guess people who are worried about filesize have the option to leave out `unicode-properties.js`.

Either way, we should account for or update the handling for `Assigned` that I highlighted above. Do you think it's the right thing to do to continue using the special `inverseOf` handling for `Assigned` rather than explicitly including all the codepoints for it in `unicode-properties.js`?

Happy to use more global lint rules (I don't think there's benefit from significant differences for `.eslintrc.js` between root, `tests`, and `tools/scripts`), and to change lint rules in whatever ways you guys think is appropriate based on what enables us to get there and based on modern ES5+ conventions.

I'm including the list of binary properties being considered for addition below, for posterity. None of these have been supported to this point.

```
ASCII_Hex_Digit
Bidi_Control
Bidi_Mirrored
Case_Ignorable
Cased
Changes_When_Casefolded
Changes_When_Casemapped
Changes_When_Lowercased
Changes_When_NFKC_Casefolded
Changes_When_Titlecased
Changes_When_Uppercased
Composition_Exclusion
Dash
Deprecated
Diacritic
Expands_On_NFC
Expands_On_NFD
Expands_On_NFKC
Expands_On_NFKD
Extender
Full_Composition_Exclusion
Grapheme_Base
Grapheme_Extend
Grapheme_Link
Hex_Digit
Hyphen
IDS_Binary_Operator
IDS_Trinary_Operator
ID_Continue
ID_Start
Ideographic
Join_Control
Logical_Order_Exception
Math
Other_Alphabetic
Other_Default_Ignorable_Code_Point
Other_Grapheme_Extend
Other_ID_Continue
Other_ID_Start
Other_Lowercase
Other_Math
Other_Uppercase
Pattern_Syntax
Pattern_White_Space
Prepended_Concatenation_Mark
Quotation_Mark
Radical
Sentence_Terminal
Soft_Dotted
Terminal_Punctuation
Unified_Ideograph
Variation_Selector
XID_Continue
XID_Start
```

josephfrazier:
In the interest of making this change more of a refactor than a functional change, I pushed b8ebf6c to revert to the pre-existing list of properties. Regarding file sizes, here's some before/after this commit:

`xregexp-all.js`:
* Before: `340K`
* After: `236K`

`xregexp-4.0.0-next.tgz` (what `npm install xregexp` downloads):
* Before: `148K`
* After: `132K`

You can see these for yourself by running the following command (NOTE: this will delete any files untracked by git):

    git clean -fxd && npm install && npm pack && du -sh xregexp*tgz && du -sh xregexp-all.js

Obviously, feel free to revert, @mathiasbynens :)

slevithan:
That works, and thanks for the file size data.

Merging this, but @mathiasbynens feel free to follow up with a PR for the other properties if you'd like.

mathiasbynens:
For the binary properties, I agree we don’t need to support all of them.

Here’s the diff between the list of properties that are supported in native JS `\p{…}` regular expressions and those in XRegExp: https://tc39.github.io/ecma262/#table-binary-unicode-properties

```diff
-ASCII_Hex_Digit
-Bidi_Control
-Bidi_Mirrored
-Case_Ignorable
-Cased
-Changes_When_Casefolded
-Changes_When_Casemapped
-Changes_When_Lowercased
-Changes_When_NFKC_Casefolded
-Changes_When_Titlecased
-Changes_When_Uppercased
-Dash
-Deprecated
-Diacritic
-Emoji
-Emoji_Component
-Emoji_Modifier
-Emoji_Modifier_Base
-Emoji_Presentation
-Extender
-Grapheme_Base
-Grapheme_Extend
-Hex_Digit
-IDS_Binary_Operator
-IDS_Trinary_Operator
-ID_Continue
-ID_Start
-Ideographic
-Join_Control
-Logical_Order_Exception
-Math
-Pattern_Syntax
-Pattern_White_Space
-Quotation_Mark
-Radical
-Regional_Indicator
-Sentence_Terminal
-Soft_Dotted
-Terminal_Punctuation
-Unified_Ideograph
-Variation_Selector
-White_Space
-XID_Continue
-XID_Start
```

Maybe we should document this difference somewhere, and refer developers to native `\p{…}` in case they need any of these properties?

slevithan:
@mathiasbynens, for reference, where do the differences in our lists come from? E.g. your list does not include the `Expands_On_*` properties, and mine didn't include the `Emoji_*` properties.

It's also interesting to see the aliases included for binary properties and scripts. XRegExp already uses its `alias` feature in `unicode-categories.js`, and could easily support e.g. `Alpha`, `DI`, `Lower`, `NChar`, `Upper`, and `space`, if these are generatable from [unicode-property-value-aliases](https://github.com/mathiasbynens/unicode-property-value-aliases).

In the future, it would be great for XRegExp to defer to native `\p{...}` support when available, rewriting the property names and prefixes as necessary to support that. I'm also very open to breaking changes that make XRegExp stricter by matching ES2019 support/rules for `\p{...}` and or making XRegExp's support for `\p{...}` a subset of what is supported by ES2019. I'd also be fine with removing XRegExp's support for Unicode blocks, to match ES2019.

mathiasbynens:
The list of properties supported in native `\p{…}` changed several times as the proposal evolved. https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/27 provides some background on a subset of these changes. `Expands_On_*` are deprecated properties, so we decided not to support them for this new (to JS RegExp) API.

>  I'm also very open to breaking changes that make XRegExp stricter by matching ES2018 support/rules for `\p{...}` and or making XRegExp's support for `\p{...}` a subset of what is supported by ES2018.

+1 I was hoping we could do this!

(Nit: `s/ES2018/ES2019/`, not that the version number really matters…)

> I'd also be fine with removing XRegExp's support for Unicode blocks, to match ES2018.

I was gonna suggest this in particular. Blocks are not a very useful grouping anyway, so it makes sense to get rid of them. Also see https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/18 where this was discussed for native `\p{…}`.

PR #221, opened 2018-02-17
Upgrade to Unicode 10

josephfrazier:
Closes https://github.com/slevithan/xregexp/issues/215

EDIT: Kudos to @mathiasbynens for doing the real work in https://github.com/slevithan/xregexp/pull/220 and https://github.com/mathiasbynens/node-unicode-data


tools/output/blocks.js
```diff
@@ -1090,5 +1090,10 @@ module.exports = [
     {
         'name': 'InYijing_Hexagram_Symbols',
         'bmp': '\u4DC0-\u4DFF'
+    },
+    {
+        'name': 'Inundefined',
```
mathiasbynens: This seems like a bug

josephfrazier: Good catch! It looks like `unicode-10.0.0` has an "undefined" block, so I'll open an issue in node-unicode-data.

josephfrazier:
Should we release v4.1.0 now that Unicode 10 is supported?

slevithan:
Yes, sounds good.

PR #226, opened 2018-02-21
Fix NPM package missing tools dir

Delagen:
Need NPM republish


josephfrazier:
Good catch, thanks for fixing it! However, it looks like @slevithan took care of it in https://github.com/slevithan/xregexp/issues/227

PR #236, opened 2018-04-18
ESLint refactoring

brettz9:
- Refactoring: Fix ESLint issue with preferring `const`; bring `let` closer to utilizing scope, destructure


josephfrazier:
Thanks again, @brettz9! I added a couple extra changes on top of yours, and I'll go ahead and merge. I didn't upgrade the syntax of any files in `tests/`, because they weren't part of [the original syntax upgrade effort](https://github.com/slevithan/xregexp/search?q=lebab&type=Commits) (I can't recall why, but maybe so that older Node versions could run the tests directly? At any rate, we can always upgrade the tests later.)

PR #237, opened 2018-04-18
ESLint: Enforce object destructuring

josephfrazier:
This is a follow-up to https://github.com/slevithan/xregexp/pull/236
We could also do array destructuring, but it looked a little
weird to me, so I left it out for now.

EDIT: Enforcing array destructuring also touches files in `tools/`, and I'm not sure if we have a reason not to do that. See https://github.com/slevithan/xregexp/pull/236#issuecomment-382510746

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -249,7 +249,7 @@ module.exports = {
             "error",
             {
                 "array": false,
-                "object": false
+                "object": true
             }
         ],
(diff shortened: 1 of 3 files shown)


slevithan:
IMO it's preferable to enforce the same style/rules across all JS code in main lib, addons, tests, and tools.

PR #243, opened 2018-04-26
Rollup

brettz9:
- Enhancement: Add Rollup ES distribution (with minification and sourcemaps)
- Enhancement: Add `module` to `package.json`
- npm: Remove browserify in favor of Rollup

As discussed in #234


brettz9:
Btw, should xregexp-all.js at least be deprecated though, given that the PR puts UMD versions in `dist` (along with the new ES6 versions)?

brettz9:
Also, I added `dist` to `package.json`'s `files` property, as `lib` and `xregexp-all.js` were already there, but if each user is getting this run in a build step during install, shouldn't none of these (`dist`, `lib`, `xregexp-all.js`) be specified there?

rollup.config.js
```diff
@@ -0,0 +1,63 @@
...
+                    ['env', {
+                        modules: false,
+                        exclude: ['transform-es2015-literals']
+                    }]
+                ],
+                plugins: [
+                    'array-includes',
+                    'transform-xregexp'
+                ]
+            }),
```
josephfrazier: Is there a way to reduce the duplication between these babel options and the `.babelrc` file? Perhaps rollup could generate the files in `lib/` instead of the `npm run babel` script that's part of `npm run prebuild`?

package.json
```diff
@@ -16,18 +16,21 @@
     "unicode"
   ],
   "main": "./lib",
+  "module": "./src/index.js",
```
josephfrazier: I think we'd probably want the module file to point to something that has passed through babel, to make sure that the only non-ES5 feature used is the import/export syntax.

rollup.config.js
brettz9: Re: one aspect of `.babelrc` duplication, I could probably just import the `.babelrc` file here and adapt it...

However, as far as Rollup replacing the script acting on `lib`, no, I don't think so. Rollup is concerned with rolling up the files into a single file whereas Babel just provides the backward-compatible code (and which Rollup itself uses as a dependency).

package.json
brettz9: Ok, sure. I had got it into my head that Rollup took advantage of having access to the source for optimizations, but I guess I was mistaken.

rollup.config.js
josephfrazier: That sounds good. It looks like we should just move the contents of `.babelrc` into the `babel` property of `package.json`. That way, we can import/require it easily from `rollup.config.js`.

https://babeljs.io/docs/usage/babelrc

brettz9: Yes, that should work, too...

brettz9:
I've updated the branch.

josephfrazier:
Thanks, it looks like the changes so far leave `lib/` as it was on the master branch, so we know that people who `npm install xregexp` and use a bundler that ignores the "module" field of package.json shouldn't be affected by this. I assume that "module"-aware-bundler users also wouldn't see any changes in behavior, given the maturity of Rollup.

However, there's one problem that I noticed with `xregexp-all.js`: It no longer sets the global `XRegExp` variable when loaded in a browser. To see this, you can open the `tests/index.html` file after running the build script. On master, this works because of browserify's `--standalone` option, so Rollup should do a similar thing for this file. Once that's fixed, I think this will be ok to merge.

Also, if you could make these and further changes as separate commits, rather than force-pushing the branch, it'll be easier to keep track of them.

Thanks!

brettz9:
Apologies--Rollup had actually been handling that aspect, but an error was preventing it from being set up. Now fixed with those HTML browser tests confirmed as running.

brettz9:
See my latest commit. Now down to 236223 (without minifying).

josephfrazier:
Thanks, I added a couple of changes to prevent unnecessary files from being published and to simplify the Rollup config. Details are in the commit messages.

brettz9:
Apologies for the delayed reply. Looks good. The only concern I have is that if you want to deprecate using xregexp-all.js (so there is consistency in using the `dist` directory), I think it'd be a good idea to keep in that non-minified build (i.e., a simple call to `getRollupObject(),`) so people can be advised to start using it.

jsg2021:
I've submitted an alternate to this PR (#282) In that PR, we use rollup to generate the `./xregexp-all.js` (umd) I made the package point `module` aware loaders (rollup/webpack) to the `./src` instead and cjs loaders (node) point to the umd... this should cut the publish size in half and simplify the package. No need to deprecate the xregexp-all, and no external change other than size 😊

brettz9:
That would work for me, but the reason I took the approach of a separate `dist` is that there would be no future breaking changes if xregexp needed to add dependencies; in such a case, `src` would no longer work out of the box, e.g., browsers could not load it as a module. And since a `dist` folder was being added, it makes sense to be able to look for the UMD build in the same place as well. If the project is confident they won't add dependencies, then this precaution isn't necessary.

jsg2021:
If the project is using non-standard features or newer syntax not in node8 then yes, we’d need to run it through babel. In which case, I’d recommend not making a dual-entry package and just publish the umd.

The rollup config can output all formats we would want. cjs/umd/esm.

brettz9:
So you would require that consumers then do their own rolling up, albeit with a provided config?

While that is better than nothing, I personally prefer the "pre-built binaries available" approach, with `module` pointing to a usable file. ESM is not only useful for projects with existing build routines, but also for demos and applications which deliberately avoid an extra build step for rapid development.

brettz9:
There is the Babel need too, but I was referring to the fact that browsers wouldn't recognize paths for dependencies you might add like `import 'some-npm-module'`. If you were importing dependencies in this manner in source, `module` would not point to a file that was usable (unless users did their own rolling up).

jsg2021:
I think you misunderstood me. I was saying the umd is the pre-built binary that serves browsers (via script tag) and commonjs. Newer projects using webpack/rollup would just work as they do now.

brettz9:
But projects which don't use Rollup and wish to use modules for modularity (avoiding polluting their global HTML with dependency chain information or their JavaScript with global variables) would not be able to do so without an ESM build.

jsg2021:
Which is why i conceded to your point several comments ago. We just have rollup produce a umd and an esm. Remove src as the module entry point.

brettz9:
Ok, I see, my apologies then. Yes, that sounds good to me.

josephfrazier:
Hey @brettz9 and @jsg2021, I'm starting to look back into using rollup. I've tried to update both this PR and https://github.com/slevithan/xregexp/pull/282 against the `master` branch, but I'm not sure one I should continue on. Do you both each prefer your own branch, or has one of you convinced the other their approach is better?

jsg2021:
If my pr addressed @brettz9 's concerns (I think it did) then you may find mine desirable since it is setup to make rollup output all your various bundle formats (umd/esm/cjs) in one go.

jsg2021:
But I don't really have a preference. Not having any dependencies on babel or core-js would be my only real desire.

brettz9:
I don't have much time to look at this currently, but I expect the other PR should be fine (though unlike per https://github.com/slevithan/xregexp/pull/282/files#issuecomment-729919943 , I do hope it will keep its ESM build so that browser demos could use the ESM build directly from `node_modules`.

The PR should probably switch to the currently maintained `@rollup/plugin-eslint` and `@rollup/plugin-babel`. I think the latter's API changes to `babelHelpers: 'bundled'`, however.

slevithan:
@josephfrazier Do you expect to include this PR or #282 in v5.0.0? There's nothing currently holding back that release, but it might be nice to include build changes at the same time.

josephfrazier:
Thanks for chiming in, @jsg2021 and @brettz9. I can't really give an estimate of when I'll be able to get this done, but I am interested in doing so.

@slevithan, I see you merged some more bug fixes, and I wouldn't want to hold up the v5.0.0 release on account of this, so I'll try to get it out today.

axljoker03 (approved):

PR #248, opened 2018-05-20
Upgrade to Unicode 11

josephfrazier:
For reference/comparison, here's the Unicode 10 upgrade:
https://github.com/slevithan/xregexp/pull/221/


mathiasbynens:
I’d prefer to wait until Unicode v11 is finalized.

josephfrazier:
Ah, thanks for the info. I don't keep track of the Unicode release schedule, but I noticed you had a the new package out, so I did a blind upgrade. I think it makes sense to wait, though.

mathiasbynens:
Unicode v11 is now released. @josephfrazier, could you please update the dependency to the latest version and run the build script again? We should be good to go then.

josephfrazier:
Done, but it looks like none of the generated files changed :man_shrugging:

josephfrazier:
Thanks for merging! @slevithan, should we release `v4.2.0`?

slevithan:
@josephfrazier, done. v4.2.0 released and published on npm.

josephfrazier:
Sweet, thanks!

Issue #262, opened 2019-01-07
Getting Error: Cannot find module '@babel/runtime/core-js/object/create'


Michaeltym:
I'm using version `4.2.2` and getting error `Error: Cannot find module '@babel/runtime/core-js/object/create'`

Here is the code (From README)
```
const XRegExp = require('xregexp');

const date = XRegExp(
  `(?<year>  [0-9]{4} ) -?  # year
   (?<month> [0-9]{2} ) -?  # month
   (?<day>   [0-9]{2} )     # day`, 'x');

let match = XRegExp.exec('2017-02-22', date);

console.log(match.year);
```

Can confirm that `4.2.0` is working

cronvel:
Same problem here.

lorenc-tomasz:
@josephfrazier The https://github.com/josephfrazier/xregexp/commit/30338174bfa5b555e200e6f9ad15f1f7af2388ba is very ugly ;)

josephfrazier:
@lorenc-tomasz 	it sure is ugly, but I think it will work for now ;) Ideally we'd upgrade to a non-beta version (looks like 7.2.0 is out now) but that might take longer

slevithan:
Thanks for the reports. And thanks for the quick fix, @josephfrazier. Should now be fixed in v4.2.3. Broken version v4.2.2 has been deprecated on npm.

josephfrazier:
Sure thing! I'll probably try to upgrade all our Babel stuff to non-beta sometime this week

PR #263, opened 2019-01-07
Pin babel runtime and transformation plugin

josephfrazier:
This should be a quick/dirty fix for https://github.com/slevithan/xregexp/issues/262

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "@babel/cli": "^7.0.0-beta.47",
     "@babel/core": "^7.0.0-beta.47",
-    "@babel/plugin-transform-runtime": "^7.0.0-beta.53",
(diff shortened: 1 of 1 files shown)


slevithan:
@josephfrazier, should I release this as v4.2.3?

josephfrazier:
I believe so, yes. I tested it by doing the build, and then cd'ing to a temporary directory, running `npm install ~/workspace/xregexp` (where my checkout is), then running the test code from #262:

```js
const XRegExp = require('xregexp');

const date = XRegExp(
  `(?<year>  [0-9]{4} ) -?  # year
   (?<month> [0-9]{2} ) -?  # month
   (?<day>   [0-9]{2} )     # day`, 'x');

let match = XRegExp.exec('2017-02-22', date);

console.log(match.year);
```

You can compare it with `npm i xregexp@4.2.2` to confirm that the test fails in that case (as reported in #262)

Issue #259, opened 2018-12-07
Exception on React Native 


lorenc-tomasz:
> undefined is not a function (evaluating 'optionalFlags[typeof Symbol === "function" ? Symbol.iterator : "@@iterator"]()')
> addToken
>     xregexp.js:760:28
> <unknown>
>     xregexp.js:2000:17
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     index.js:8:45
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     usernameValidator.js:2
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     ProfileHomeScreen.js:24
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     HomeStackNavigator.js:8
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     App.js:46
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:214:36
> metroRequire
>     require.js:136:24
> <unknown>
>     index.android.js:3
> loadModuleImplementation
>     require.js:320:6
> guardedLoadModule
>     require.js:207:45
> metroRequire
>     require.js:136:24
> global code
>     
> 

With 3.2.0 everything works ok

ducaale:
version 4.1.1 also works

josephfrazier:
@slevithan, it looks like this might be due to the use of `Symbol`, which also caused problems in https://github.com/slevithan/xregexp/issues/254, and was fixed after version `4.2.0`. Should we release version `4.2.1` and see if it helps?

josephfrazier:
Great, I'll put together the PR

lorenc-tomasz:
4.2.2:

`Unable to resolve module `@babel/runtime/core-js/get-iterator` from `/Projects/xyz/node_modules/xregexp/lib/addons/unicode-base.js`: Module `@babel/runtime/core-js/get-iterator` does not exist in the Haste module map`

josephfrazier:
[...]

Issue #254, opened 2018-07-13
Use of Symbol breaks IE and older browser support


onexdata:
I didn't see this issue in 4.1.1, but it looks like 4.2.0 uses Symbol, which is certainly not compatible with the older browsers listed on regexp.com (Internet Explorer 5.5+, Firefox 1.5+, Chrome, Safari 3+, and Opera 11+)

slevithan:
xregexp.com is currently out of date and only lists pre-v4 versions at <http://xregexp.com/history/>. This is on me to upgrade since the documentation is not currently open source. :'( I aim to fix this but am not able to provide a timeline at this time.

XRegExp moved to only support ES5+ browsers in v4.0.0.

Looks like use of `Symbol.iterator` in xregexp-all.js came from @josephfrazier's https://github.com/slevithan/xregexp/pull/236/commits/8c02a0a4e5700a099e7c32a9437af2cedb17f5c7 and is indeed new in v4.2.0. This might have been unintentional. @josephfrazier, what do you think is the best path forward for this?

josephfrazier:
Thanks for tracking down that commit, it looks like https://github.com/slevithan/xregexp/pull/238 also had changes which introduced `Symbol` to xregexp-all.js. See #255 for a potential fix.

PR #275, opened 2019-12-19
feat(babel): remove babel helper from dependencies as core-js2 outdated

Delagen:


jsg2021:
This would be a very welcome change. I'm currently pinning to 4.2.0 so I don't incur the additional `@babel/runtime-corejs2` dep

slevithan:
@josephfrazier, is this the right fix?

josephfrazier:
I'm not sure, I need to go back and see why we're using the helpers in the first place. If I can get https://github.com/slevithan/xregexp/pull/280 working, it should remove the deprecation, without potentially breaking anything.

josephfrazier:
I did some `git blame .babelrc` and `git log -S '@babel/plugin-transform-runtime'` and found that `@babel/plugin-transform-runtime` was added in https://github.com/slevithan/xregexp/pull/255 to fix IE compatibility (https://github.com/slevithan/xregexp/issues/254), so I wouldn't recommend removing it at the moment. If we had more comprehensive browser testing (perhaps based on market share), we could get a better idea of what XRegExp browser compatibility looks like and how PRs like this would affect it.

jsg2021:
That doesn’t make sense.  The runtime is just sharing shortcuts/helpers
that babel used to replace es6+ syntax. It’s not a polyfill. If you
configure babel to use builtins or turn them off... your code is no more or
less compatible with IE.

Honestly, the best solution would be to ship your esm code as-is, in
parallel, under a module entry. So those of us using it with
esm/webpack/rollup wouldn’t hit the transpiled code at all. (you still
should not depend on the babel runtime if you can help it)

Delagen:
@josephfrazier It does not remove IE support, it just include some parts of runtime in code, so not runtime needed at all.

jsg2021:
The current release make this a little less painful but I think the babel runtime should not be a dep. This change would accomplish that. However, #282 would also accomplish this AND simplify this package some. Legacy commonjs and script-tag consumers get a self-contained bundle and module-aware consumers get the original source.

josephfrazier:
> @josephfrazier It does not remove IE support, it just include some parts of runtime in code, so no runtime needed at all.

If IE does not support `Symbol`, then I believe the runtime is needed. See https://github.com/slevithan/xregexp/issues/259#issuecomment-450537120, where a Symbol-related issue was fixed by adding the runtime.

Sorry for the delay here, btw. It's been a crazy year.

jsg2021:
babel-runtime doesn't add Symbols. If you target IE, babel will compile symbol references and symbol consuming structures into using core-js' apis. This means you will depend on core-js. 

Honestly, imo, you should probably drop all this and just publish your code as modern esm (or commonjs). Make your consumers responsible for compiling down to legacy targets. 

You can still provide a umd script that's pre-built and bundled ... the tools and deps needed for that can be devDeps... that won't require external dependencies.

axljoker03 (approved):

axljoker03 (approved):

PR #281, opened 2020-01-31
Add `module` to package.json for rollup/webpack

jsg2021:
This will allow users to consume the original esm modern code. This would make #275 a minor non-blocking issue. :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "regular expression",
     "unicode"
   ],
-  "main": "./lib",
+  "main": "./lib/index.js",
+  "module": "./src/index.js",
   "files": [
     "src",
     "lib",


jsg2021:
I rebased onto master

josephfrazier:
Thanks! I still need a bit more time to evaluate https://github.com/slevithan/xregexp/pull/282 and https://github.com/slevithan/xregexp/pull/243, but this looks like a no-brainer.

josephfrazier:
Hmm actually, after clicking through those PRs, I found this comment at https://github.com/slevithan/xregexp/pull/243#discussion_r186300617:

> I think we'd probably want the module file to point to something that has passed through babel, to make sure that the only non-ES5 feature used is the import/export syntax.
> 
> EDIT: See [rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features)

I guess we can see if this is actually a problem, once this change is published.

jsg2021:
Ah well, if there are usages of non-standard syntax or features not in node 8, then we’d just remove this. 

The only reason I proposed it was to prevent bundlers from importing (yet another version of/potentially conflicting) babel runtime. As well as potentially making the bundled size even smaller. 

The real and final solution will be to reconfigure babel not to use the runtime. The Rollup change does this. (The runtime is meant to be used by applications to minimize babel helper code duplication at the app level... libraries are discouraged from using it)

jsg2021:
I've updated my other pr to produce a esm module so that the src goes through babel.

jantimon:
@jsg2021 this change broke xregexp for all webpack users - please see #305

jsg2021:
All webpack users? or commonjs-webpack users?

jantimon:
for all webpack users

jsg2021:
Ah

PR #282, opened 2020-01-31
Use rollup to simplify build/babel/lint

jsg2021:
This updates all devDependencies to latest (as of now) and removes the need for any babel-runtime dependencies (rollup bakes them in & dedups them for us) This also moves `zuul` to a npx (on demand) script (since it prints many deprecation warnings atm) The only deprecation warning on install is core-js@2 (which comes from a babel-transform plugin)

Thoughts?

With this change, rollup produces both cjs & umd bundles. `npm test` passes.


jsg2021:
I just noticed #243 ... this maybe a cleaner approach. I made this produce the cjs/umd instead of producing an es module... This makes the original `src/` the es code...and the cjs/umd the "fallbacks".

jsg2021:
since the umd is nearly ident to the cjs, I could just make the `main` point to the umd. and stop producing the `/lib/index.js`... that way there is only one fallback that covers browser & node (cjs).

jsg2021:
I've rebased onto your master.

jsg2021:
I've updated this to incorporate yesterdays comments from #243 ... I've also fully commented the rollup config.

jsg2021:
Squashed & rebased. Hope this is found usefull :)

slevithan:
This sounds good, but I'm not actively supporting XRegExp these days and am not actually that familiar with cjs vs umd vs esm, rollup, babel, etc. so I'd appreciate insight from @josephfrazier or others.

jsg2021:
This should probably be inverted to produce a commonjs & umd build instead of esm & umd... let modern node have the raw sources.

Issue #284, opened 2020-02-06
include TypeScript definitions


tdjastrzebski:
Please include TypeScript definition with the upcoming new package version.

PR #285, opened 2020-02-11
GH-284: Add TypeScript definitions and tests

jfahrenkrug:
Ported TS definitions from DefinitelyTyped, including tests.

Once a version with these definitions is released, we will need
to remove the definitions from DefinitelyTyped:
https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2180,6 +2180,12 @@
         }
       }
     },
+    "@types/parsimmon": {
+      "version": "1.10.1",
+      "resolved": "https://registry.npmjs.org/@types/parsimmon/-/parsimmon-1.10.1.tgz",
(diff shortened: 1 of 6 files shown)


jfahrenkrug:
@josephfrazier Good catch! I didn't do a simple port though. I used https://github.com/Microsoft/dts-gen to create a baseline, then manually checked the JS methods to make sure the types are correct (dts-gen uses `any` a lot). Please let me know if you want to have any improvements to the code!

slevithan:
@jfahrenkrug This may be a lot to ask, but any ideas why `dtslint` v3.0.0+ (from 2 years ago, released a few days after this pull request) is failing for XRegExp? New versions of `dtslint` fail with error "TypeError: Cannot read property 'NameOnly' of undefined", but there are no references to "NameOnly" anywhere in the XRegExp codebase and I haven't been able to find information about `dtslint` v3 breaking changes.

jfahrenkrug:
@slevithan Hey Steven, I have no idea. I'd be happy to take a look, but I'm not sure if I'll find time this week.

PR #286, opened 2020-03-10
Add non-nullable modifier to return type of functions never returning null

vrana:
Alternatively use primitive type instead of nullable boxed type.


josephfrazier:
I'm not very familiar with [JSDoc types](https://jsdoc.app/tags-type.html), but this seems fine to me. Thanks!

PR #288, opened 2020-03-16
Updated types def and tests

pristinesource:


pristinesource:
Created here from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43115 as per request from @jfahrenkrug

josephfrazier:
Thanks, especially for the link to the DefinitelyTyped PR! I eyeballed this one against as a sanity check, and it seems fine. I'm hoping to be able to make a new release sometime this week.

josephfrazier:
> I'm hoping to be able to make a new release sometime this week.

Hmm actually, it looks like I no longer have NPM public permissions, according to  `npm owner ls xregexp`, so we might have to wait for @slevithan if we want to get  a new version out, containing [the changes since v4.3.0](https://github.com/slevithan/xregexp/compare/v4.3.0...master).

slevithan:
@josephfrazier I haven’t changed anything so I think you should still be able to publish. I just checked the `xregexp` package settings on npmjs.com and you have write access.

PR #299, opened 2020-09-07
Added babel plugin and loop export

Piterden:
Added plugin proposal export namespace from. And shortened some exports.


josephfrazier:
Thanks for the PR, @Piterden! I'm a little hesitant to merge it, though, as the proposal is currently only at Stage 1: https://github.com/tc39/proposal-export-default-from, and I'm not sure if it's worth the same amount of code cleanup it allows

slevithan:
Agree with @josephfrazier. I don't think this adds enough improvement to justify the additional dependency, especially if the TC39 proposal is not on a clear track to approval. I'm going to close this, but we can always reopen later if we made the wrong call here. Thanks for the PR though, @Piterden -- it's interesting to see.

Issue #305, opened 2020-11-04
Import does not work anymore since version 4.4.0


gizmodus:
We use xregexp in a shared library that is used in our client (Angular 10) as well as on our server (NodeJS / Webpack 5). Until version 4.3.0 we were able to import xregexp like this and everything worked as expected:

`const XRegExp = require('xregexp');`

Now we wanted to upgrade to version 4.4.0, but get the following error in the client application (Angular 10):

`TypeError: XRegExp.matchRecursive is not a function`

The server still works as expected.

Changing the import statement to `const XRegExp = require('xregexp').default;` solves the problem in the client but leads to this error on the server: `TypeError: Cannot read property 'matchRecursive' of undefined`.

We also tried `import * as XRegExp from 'xregexp';` but this does not work either. Any ideas why this suddenly does not work anymore?

sachithd:
Facing the same issue.

slevithan:
Possibly this is an unfortunate side effect of upgrading most all of the dependencies in this release. Alternatively this could have stemmed from adding TypeScript definitions. Any pull requests that fix this are most welcome.

josephfrazier:
I suspect this may be an Angular issue, see https://github.com/slevithan/xregexp/issues/294

jsg2021:
If you're using webpack and not using esm, you should probably configure webpack to ignore `module` main fields since you are going against the grain.

https://webpack.js.org/configuration/resolve/#resolvemainfields

jsg2021:
For the OP, use:

```
import XRegExp from 'xregexp';
```

jantimon:
@jsg2021 we are using a lot of es6 modules node_modules for better tree shaking so disabling the `module` field isn't an option for us

right now we are using the @adobe/htlengine which is using xregexp somewhere in its xss validaiton logic https://github.com/adobe/htlengine/blob/ffcfeecc1d60e0f8f47c15be68ce794a8620b6d6/src/runtime/xss_api.js

your change broke our entire pipeline :)

jsg2021:
My PR is more than 9mo old and I think XRegExp was still on 3.x...

To your point, the evidence says this change is worthy of a major version bump.

jantimon:
4.4.0 was released 20 days ago and this issue was filed 14 days ago

xregexp 4.3.0 has no `module` field:
https://unpkg.com/xregexp@4.3.0/package.json  

xregexp 4.4.0 has a `module` field:
[...]

PR #308, opened 2020-12-08
Add browser field to package.json to fix webpack

josephfrazier:
This should close https://github.com/slevithan/xregexp/issues/305, see https://github.com/slevithan/xregexp/issues/305#issuecomment-740353997:

> Actually, after reading https://webpack.js.org/configuration/resolve/#resolvemainfields more closely, it sounds like adding a `browser` field to `package.json`, that points to `lib/index.js` (same as `main`) should fix things.
>
> > resolve.mainFields
> > [string]
> >
> > When importing from an npm package, e.g. import * as D3 from 'd3', this option will determine which fields in its package.json are checked. The default values will vary based upon the target specified in your webpack configuration.
> >
> > When the target property is set to webworker, web, or left unspecified:
> >
> > webpack.config.js
> >
> > module.exports = {
> >   //...
> >   resolve: {
> >     mainFields: ['browser', 'module', 'main']
> >   }
> > };
> > For any other target (including node):
> >
> > webpack.config.js
> >
> > module.exports = {
> >   //...
> >   resolve: {
> >     mainFields: ['module', 'main']
> >   }
> > };
> > For example, consider an arbitrary library called upstream with a package.json that contains the following fields:
> >
> > {
> >   "browser": "build/upstream.js",
> >   "module": "index"
> > }
> > When we import * as Upstream from 'upstream' this will actually resolve to the file in the browser property. The browser property takes precedence because it's the first item in mainFields. Meanwhile, a Node.js application bundled by webpack will first try to resolve using the file in the module field.
>
> I'll open a PR for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "regular expression",
     "unicode"
   ],
+  "browser": "./lib/index.js",
   "main": "./lib/index.js",
   "module": "./src/index.js",
   "files": [


slevithan:
Nice research. Based on the docs you linked to, this looks like it will indeed resolve the issue for anyone with webpack's `target` property set to `webworker`, `web`, or left unspecified.

@josephfrazier, do you think this should be released as a 4.4.1 or 4.5.0?

josephfrazier:
Thanks! This feels like a 4.4.1 bugfix to me

josephfrazier:
@slevithan are you going to take care of doing the release, or should I?

slevithan:
@josephfrazier If you're okay with doing it, I'd really appreciate it! I'm trying to minimize involvement going forward :( But let me know if you run into any problems and I'll be happy to help!

josephfrazier:
I should be able to dig up and do all the steps, especially now that you've made it easier with https://github.com/slevithan/xregexp/commit/760ab95bcef8948f313df55e9483981200b70e2d

I'll give it a shot and reach out if there's any issues!

josephfrazier:
@slevithan can I get collaborator access to this repo (https://github.com/slevithan/xregexp/settings/access)?

Specifically, I'd like to add a branch protection rule that requires the new github action check to pass on PRs: https://github.com/josephfrazier/xregexp/settings/branches

slevithan:
@josephfrazier You already had collaborator access, so I'm not sure what's up. I just removed then re-invited you.

josephfrazier:
Ohh, whoops. I think I meant to ask for _maintain_ or _admin_ access, so I can manage the repo settings at https://github.com/slevithan/xregexp/settings. Here's what it currently looks like for me:

image

slevithan:
@josephfrazier I don't see any option to change collaborators' roles/permissions. I'm guessing that's because this is a personal repo, rather than owned by an organization.

I've gone ahead and created a branch protection rule. I set the branch name pattern as `master` and checked only the option "Require status checks to pass before merging" (with "build" as the selected status check). Let me know if you want me to make any further changes!

(If you continue to need admin access in the future, I'm open to discussing transferring the repo to an organization.)

josephfrazier:
Thanks, that looks like it did the trick! For an example, see https://github.com/slevithan/xregexp/pull/312

I don't know if we need to bother with an organization yet, I'll see if I hit future barriers

PR #247, opened 2018-05-06
BREAKING: Handle ES2018 capture names

josephfrazier:
See https://github.com/slevithan/xregexp/commit/1050835ccb31fc370373994a1b09026bdc29b39b
and https://github.com/slevithan/xregexp/pull/241#issuecomment-384480732

I think this should close https://github.com/slevithan/xregexp/issues/224

~# TODO:~

* [x] Update readme and any other docs in the repo
* [x] Remove dead code

PR #315, opened 2021-01-09
Use Istanbul to show code coverage at coverage/lcov-report/index.html

josephfrazier:
It also prints it out after the tests like this:

```
=============================== Coverage summary ===============================
Statements   : 74.02% ( 1946/2629 )
Branches     : 54.6% ( 878/1608 )
Functions    : 75.5% ( 413/547 )
Lines        : 78.67% ( 1870/2377 )
================================================================================
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 lib
 xregexp-all.js
+.nyc_output
+coverage
 
 # Compiled Python files
 *.pyc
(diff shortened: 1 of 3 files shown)


slevithan:
Since review was requested.... I'm down with this if you think it's useful. :)

(Stating the obvious, this won't give insight into the subtleties of regex and replacement text syntax that need to be tested.)

josephfrazier:
Thanks for taking a look. My intent was to find code that would be made dead by https://github.com/slevithan/xregexp/pull/247 so it could be removed, but it didn't actually change the coverage of many lines.

I'll go ahead and merge it, since it does provide some insight into which code gets used by the test suite, at least one a given machine (I think a lot of the uncovered branches have to do with legacy support, so wouldn't get run on my machine).

PR #316, opened 2021-01-09
BREAKING: Enable `namespacing` feature by default

josephfrazier:
Following up on the [change] to use ES2018 constraints on the names of capture groups,
this makes it so that XRegExp `namespacing` feature is enabled by
default, so that named capture group matches will appear on the
`.groups` property of the match object, rather than directly on the
match object, in accordance with the ES2018 spec.

While this is a breaking change, users can restore the old behavior by running:

    XRegExp.uninstall('namespacing')

[change]: https://github.com/slevithan/xregexp/pull/247#issuecomment-757374810

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ const date = XRegExp(
      (?<month> [0-9]{2} ) -?  # month
      (?<day>   [0-9]{2} )     # day`, 'x');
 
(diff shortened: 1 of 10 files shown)


slevithan (approved):
Looks really solid!

josephfrazier:
thanks!

PR #332, opened 2021-07-29
Improve XRegExp.matchRecursive handling of unbalanced delimiters (Clo…

dvargas92495:
…ses #96)

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,18 +224,18 @@ const str4 = '<1><<<2>>><3>4<5>';
 XRegExp.matchRecursive(str4, '<', '>', 'gy');
 // -> ['1', '<<2>>', '3']
 
-// Returning unmatched delimiters as raw text instead of erroring
-const str5 = 'Here is <div> <div>an</div> unmatched example';
+// Skipping unbalanced delimiters instead of erroring
+const str5 = 'Here is <div> <div>an</div> unbalanced example';
 XRegExp.matchRecursive(str5, '<div\\s*>', '</div>', 'gi', {
     valueNames: ['between', 'left', 'match', 'right'],
-    unbalancedDelimiters: 'text',
+    unbalancedDelimiters: 'skip',
 });
 /* -> [
-{name: 'between', value: 'Here is <div> ',     start: 0,  end: 14},
-{name: 'left',    value: '<div>',              start: 14, end: 19},
-{name: 'match',   value: 'an',                 start: 19, end: 21},
-{name: 'right',   value: '</div>',             start: 21, end: 27},
-{name: 'between', value: ' unmatched example', start: 27, end: 45}
+{name: 'between', value: 'Here is <div> ',      start: 0,  end: 14},
+{name: 'left',    value: '<div>',               start: 14, end: 19},
+{name: 'match',   value: 'an',                  start: 19, end: 21},
+{name: 'right',   value: '</div>',              start: 21, end: 27},
+{name: 'between', value: ' unbalanced example', start: 27, end: 44}
 ] */
 ```
 
--- a/README.md
+++ b/README.md
@@ -231,7 +231,11 @@ XRegExp.matchRecursive(str5, '<div\\s*>', '</div>', 'gi', {
     unbalancedDelimiters: 'text',
 });
 /* -> [
-
+{name: 'between', value: 'Here is <div> ',     start: 0,  end: 14},
+{name: 'left',    value: '<div>',              start: 14, end: 19},
+{name: 'match',   value: 'an',                 start: 19, end: 21},
+{name: 'right',   value: '</div>',             start: 21, end: 27},
+{name: 'between', value: ' unmatched example', start: 27, end: 45}
 ] */
 ```
 
--- a/README.md
+++ b/README.md
@@ -223,6 +223,16 @@ XRegExp.matchRecursive(str3, '{', '}', 'g', {
 const str4 = '<1><<<2>>><3>4<5>';
 XRegExp.matchRecursive(str4, '<', '>', 'gy');
 // -> ['1', '<<2>>', '3']
+
+// Returning unmatched delimiters as raw text instead of erroring
+const str5 = 'Here is <div> <div>an</div> unmatched example';
+XRegExp.matchRecursive(str5, '<div\\s*>', '</div>', 'gi', {
+    valueNames: ['between', 'left', 'match', 'right'],
+    unbalancedDelimiters: 'text',
+});
+/* -> [
+
+] */
 ```
 
 `XRegExp.matchRecursive` throws an error if it scans past an unbalanced delimiter in the target string.
@@ -246,6 +256,15 @@ In [Node.js](https://nodejs.org/en/):
 ```js
 const XRegExp = require('xregexp');
 ```
+
+## Contribution Guide
+
+1. Fork the repository and clone the forked version locally.
+1. Ensure you have the `typescript` module installed globally.
+1. Run `npm install`.
+1. Ensure all tests pass with `npm t`.
+1. Add tests for new functionality or that fail from the bug not fixed.
+1. Implement functionality or bug fix to pass the test.
 
 ## Credits
 
--- a/src/addons/matchrecursive.js
+++ b/src/addons/matchrecursive.js
@@ -30,7 +30,7 @@ export default (XRegExp) => {
      * @param {String} left Left delimiter as an XRegExp pattern.
      * @param {String} right Right delimiter as an XRegExp pattern.
      * @param {String} [flags] Any combination of XRegExp flags, used for the left and right delimiters.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -247,6 +257,15 @@ In [Node.js](https://nodejs.org/en/):
...
 
+## Contribution Guide
+
+1. Fork the repository and clone the forked version locally.
+1. Ensure you have the `typescript` module installed globally.
+1. Run `npm install`.
+1. Ensure all tests pass with `npm t`.
+1. Add tests for new functionality or that fail from the bug not fixed.
+1. Implement functionality or bug fix to pass the test.
+
```
dvargas92495, line 272: Added this for two reasons:
- I generally find contribution guides helpful for new repos
- Wanted to highlight how `npm install` didn't work out of the box for me as it assumed I had `typescript` installed globally.

Happy to remove it since it's out of scope. But wanted to point it out in case it's something you guys wanted to add in the near future

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
...
                     }
                 }
-            } else {
+            } else if (unbalancedDelimiters === 'error') {
                 const delimSide = rightMatch ? 'right' : 'left';
                 const errorPos = rightMatch ? delimStart : outerStart;
                 throw new Error(`Unbalanced ${delimSide} delimiter found in string at position ${errorPos}`);
+            } else if (unbalancedDelimiters === 'text') {
+                if (rightMatch) {
+                    rightMatch = null;
```
dvargas92495, line 189: this is the case when it's an unbalanced right delimiter. Simply setting rightMatch to null will treat the character as raw text since all of the requisite indices were updated

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
...
+                } else {
+                    delimEnd = XRegExp.exec(str, left, outerStart)[0].length + outerStart;
+                    openTokens = 0;
```
dvargas92495, line 192: this is the case when it's an unbalanced left delimiter. We move `delimEnd` back to right after the unbalanced left delimiter and set `openTokens` to 0 to "restart" the loop and unbalanced delimiter then is treated as raw text

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
...
+                }
+            } else {
+                throw new Error(`Unsupported value for unbalancedDelimiters: ${unbalancedDelimiters}`);
```
dvargas92495, line 195: I opted for the design choice of a string "enum" so that there could be room for various other modes in the future ("error left", "error right", "warning", "trim", etc.)

slevithan:
Haven't looked at this closely yet but just wanted to say that the general direction you took seems really solid -- thanks!

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
...
-            } else {
+            } else if (unbalancedDelimiters === 'error') {
                 const delimSide = rightMatch ? 'right' : 'left';
                 const errorPos = rightMatch ? delimStart : outerStart;
                 throw new Error(`Unbalanced ${delimSide} delimiter found in string at position ${errorPos}`);
+            } else if (unbalancedDelimiters === 'text') {
+                if (rightMatch) {
+                    rightMatch = null;
+                } else {
+                    delimEnd = XRegExp.exec(str, left, outerStart)[0].length + outerStart;
```
slevithan, line 191: Note the edge case where `left` matches an empty string. Is this already covered by the empty string delimiter handling below (`if (delimStart === delimEnd) {...}`)? If not, it needs to be covered here (`delimEnd = outerStart + (<...>.length || 1)`).

---
I think this is the correct handling -- specifically, advancing to the end of the outer left delimiter instead of advancing by one character (with a more simple `delimEnd = outerStart + 1`). But it's a choice we're making.

Consider how this can change behavior when the left delimiter is able to match at least two characters. For a simplified example, I'll use the subject string `'<<<<>>'`, left delimiter `'<<'`, and right delimiter `'>>'`, but keep in mind that the regex delimiters could be much more dynamic in what they match.

By advancing here to the end of the left delimiter, you get a _between_ value of `'<<'`, _left_ `'<<'`, _match_ `''`, and _right_ `'>>'`. However, if you were to advance by one character instead, you would get a _between_ value of `'<'`, _left_ `'<<'`, _match_ `'<'`, and _right_ `'>>'`.

I suspect your current handling will be more intuitive in all but extremely edge cases, so I think it's probably the right way to go by default. But it opens up the option of a future handling mode (I'll call it `'skip-lazy'` for now) that advances by one character when it finds an unbalanced delimiter.

src/addons/matchrecursive.js
slevithan: Great design choice.

tests/spec/s-addons-matchrecursive.js
```diff
@@ -47,6 +47,20 @@ describe('XRegExp.matchRecursive addon:', function() {
             expect(XRegExp.matchRecursive(str, '<', '>', 'gy')).toEqual(['1', '<<2>>', '3']);
         });
 
+        it('should pass the readme example for unbalanced delimiters', function() {
+            const str = 'Here is <div> <div>an</div> unmatched example';
```
slevithan, line 51: replace 'unmatched' with 'unbalanced'

slevithan (changes requested):
Fantastic work. A few small things to address.

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
                         break;
                     }
                 }
-            } else {
+            } else if (unbalancedDelimiters === 'error') {
```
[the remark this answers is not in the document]
dvargas92495: totally agree

src/addons/matchrecursive.js
```diff
@@ -179,10 +180,19 @@ export default (XRegExp) => {
...
                 const delimSide = rightMatch ? 'right' : 'left';
                 const errorPos = rightMatch ? delimStart : outerStart;
                 throw new Error(`Unbalanced ${delimSide} delimiter found in string at position ${errorPos}`);
+            } else if (unbalancedDelimiters === 'text') {
```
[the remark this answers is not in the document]
dvargas92495: not at all attached to the naming. figured I could pick something simple to convey the functionality and have you guys reply with your preference 😄. Will switch to `skip`

src/addons/matchrecursive.js
dvargas92495: ooh that's interesting. I'll note that the implementation is heavily biased towards my intended use case 😄.

Here's another edge case that this made me think of: `<<<>>`

Intuitively, one might expect a match on the last four characters. however, I believe the current implementation would skip over the whole string entirely bc outer left would increment to the substring `<>>`.  This would also be handled by your `skip-lazy` idea, but just wanted to make note of it

slevithan: Yeah, that's a great additional example. I wonder if it would be better to include both `'skip'` and `'skip-lazy'` modes right off the bat to give people the choice.

src/addons/matchrecursive.js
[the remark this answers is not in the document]
dvargas92495: This will disobey the `no-lonely-if` eslint rule. would you prefer:
- disabling for line
- removing the rule
- reverting back to else-if chain

src/addons/matchrecursive.js
dvargas92495: Turns out my example `<<<>>` was treated the same in both cases, with the middle `<` as the match. Your example showcased the difference so I used that one for the test

src/addons/matchrecursive.js
```diff
@@ -180,9 +195,23 @@ export default (XRegExp) => {
...
+                    const delimSide = rightMatch ? 'right' : 'left';
+                    const errorPos = rightMatch ? delimStart : outerStart;
+                    throw new Error(`Unbalanced ${delimSide} delimiter found in string at position ${errorPos}`);
+                } else if (unbalancedDelimiters.startsWith('skip')) {
+                    if (rightMatch) {
+                        rightMatch = null;
+                    } else {
+                        delimEnd = unbalancedDelimiters === 'skip-lazy' ?
+                            1 :
+                            (XRegExp.exec(str, left, outerStart)[0].length + outerStart) || 1;
```
slevithan, line 209: Needs to be `(XRegExp.exec(str, left, outerStart)[0].length || 1) + outerStart`, otherwise it won't work correctly if `outerStart > 0`. Will fix in a follow up.

dvargas92495: agh yea you're right, my bad!

PR #364, opened 2026-08-05
fix: replace __proto__ with Object.setPrototypeOf

titanism:
`__proto__` is deprecated (Annex B) and breaks under Node.js `--disable-proto=delete` / `--disable-proto=throw`.

Under `--disable-proto=delete`, `regex.__proto__` is `undefined` so the truthiness check fails and falls through to the manual property-copy fallback, breaking `instanceof XRegExp`.

This replaces the `__proto__` check/assignment with `Object.setPrototypeOf` (ES2015+, Node 4+). The fallback branch is preserved for pre-ES2015 engines.


slevithan:
Thanks for the PR!

titanism:
np can you release new version to npm?

titanism:
@slevithan can you release new version to npm?

slevithan:
@titanism Yes, I will do it within the next few days. Apologies for the delay on this.

titanism:
Just pinging here @slevithan to see if you can release to npm

slevithan:
@titanism I've finally published it, as v5.1.3.

Issue #362, opened 2025-01-25
Prototype Pollution Vulneralbility in xregexp v5.1.1


Ducky97:
### Affected Package:
xregexp v5.1.1

### Vulnerability Location(s): 
[function XRegExp.cache
](https://github.com/slevithan/xregexp/blob/master/src/xregexp.js#L764)

### Description: 
The latest version of xregexp (5.1.1) is vulnerable to Prototype Pollution through the entry function XRegExp.cache. An attacker can supply a payload with a pattern and flag to introduce or modify properties within the global prototype chain. 
The consequences of this vulnerability can escalate to other injection-based attacks, depending on how the library is integrated within the application. For example, if the polluted property propagates to sensitive Node.js APIs (e.g., exec, eval), it could enable an attacker to execute arbitrary commands within the application's context.

### PoC:
```javascript
(async () => {
  var victim = {}
  const x = require("xregexp")
  x.cache('/a/', 'gi');
  console.log("Before Attack: ", {}.gi); // undefined
  x.cache('__proto__', 'gi');
  console.log("After Attack: ", {}.gi); // /__proto__/gi { … }
})();
```

slevithan:
Thanks for the detailed report! Will fix.

Until shown otherwise, though, I suspect this remains in the realm of bug rather than vulnerability, due in part to the limited number of flags supported by XRegExp.

Ducky97:
Thank you for the quick response and for committing to address this issue!

While I understand your initial classification of this as a bug due to the limited number of flags supported by XRegExp, I believe this issue merits consideration as a security vulnerability for a few reasons:

1. Nature of Prototype Pollution: While this vulnerability currently only affects `__proto__` properties that match specific regex flags such as `gi`, it does alter properties on the prototype chain. This could unintentionally affect applications using XRegExp, especially those that heavily rely on object properties.
2. Potential Security Risks: If the polluted properties are used as critical control points in an application, such as for permissions checks or execution controls, this seemingly limited vulnerability could be exploited to bypass security mechanisms, potentially leading to remote code execution (RCE).
3. Raising Community Awareness: Classifying this issue as a security vulnerability would heighten awareness among developers about potential risks when implementing and using XRegExp, encouraging more robust security practices.

Given these points, even though the direct impact might appear limited, the potential for serious consequences under certain conditions suggests that it would be prudent to officially document and track this as a security vulnerability.

I look forward to your thoughts on this matter! Thank you once again!

PR #366, opened 2026-09-04
Reject `__proto__`-like flag strings that bypass the unknown-flag check (fixes #365)

### LINKED ISSUE #365: '__proto__' / 'constructor' flags silently skip Unknown regex flag validation and return an always-matching regex (follow-up to #362)
### Affected version

- v5.1.2 (latest release; npm `latest`)
- Also present on current `master` (checked at `e2fd0274`, 2026-09-02)

### Summary

`XRegExp(pattern, flags)` and `XRegExp.cache(pattern, flags)` validate the flags string via `prepareFlags()` and throw `SyntaxError: Unknown regex flag` for unsupported flags. However, when `flags` is exactly `'__proto__'` or `'constructor'`, validation is skipped and a surprising value is returned instead of an error:

- `XRegExp(pattern, '__proto__')` / `XRegExp(pattern, 'constructor')` → returns `/(?:)/`, an **always-matching** empty regex
- `XRegExp.cache(pattern, '__proto__')` → returns `Object.prototype` (not a RegExp; no `.test` method)
- `XRegExp.cache(pattern, 'constructor')` → returns the `Object` constructor function

This looks like a follow-up to #362: commit 838ce761 fixed the outer caches (`patternCache` / `regexCache` are now `Object.create(null)`), but the **inner** per-pattern cache objects are still plain `{}`, and the cache lookup happens **before** flags validation.

### Reproduction

```js
import XRegExp from 'xregexp'; // or require

XRegExp('abc', 'z');
// SyntaxError: Unknown regex flag z   ← expected behavior for unknown flags ✓

XRegExp('^admin$', '__proto__').test('guest');   // true  ← should throw SyntaxError
XRegExp('^admin$', 'constructor').test('guest'); // true  ← should throw SyntaxError

XRegExp.cache('abc', '__proto__');   // Object.prototype  — no .test method
XRegExp.cache('abc', 'constructor'); // [Function: Object]
```

Verified with Node 20 and Node 24, on both the npm `5.1.2` tarball and `master` source.

### Root cause

In `src/xregexp.js`, the constructor creates an inner plain object per pattern and looks up the flags key *before* validating the flags:

```js
let regexCache = Object.create(null);    // outer cache — fixed by #362
let patternCache = Object.create(null);  // outer cache — fixed by #362

function XRegExp(pattern, flags) {
    // ...
    if (!patternCache[pattern]) {
        patternCache[pattern] = {};                     // ← inner object is a plain {}
    }

    if (!patternCache[pattern][flags]) {                // ← lookup happens BEFORE validation
        // ... token translation ...
        const applied = prepareFlags(pattern, flags);   // ← flags validated only on cache miss
        // ...
        patternCache[pattern][flags] = { /* compiled details */ };
    }

    const generated = patternCache[pattern][flags];     // ← Object.prototype for '__proto__'
    return augment(
        new RegExp(generated.pattern, generated.flags), // undefined/undefined → /(?:)/
        // ...
    );
}
```

When `flags === '__proto__'`, `patternCache[pattern]['__proto__']` resolves through the prototype chain to `Object.prototype` (truthy), so the compile-and-validate block is skipped entirely. `generated` then *is* `Object.prototype`; `generated.pattern` and `generated.flags` are `undefined`, and `new RegExp(undefined, undefined)` produces `/(?:)/`.

`'constructor'` behaves the same way (`patternCache[pattern].constructor` is the `Object` constructor, truthy).

`XRegExp.cache` has the same shape — inner `regexCache[pattern] = {}` plus a truthiness check on `regexCache[pattern][flags]` — and there the truthy prototype value is returned directly to the caller, which is why `XRegExp.cache('abc', '__proto__')` returns `Object.prototype` itself.

### Impact

XRegExp itself doesn't treat flags as untrusted input, so I'd frame this as a validation/robustness bug rather than a direct vulnerability. But the intended guard is silently inverted for these two flag strings: any caller that passes a flags value through from external input gets an always-matching regex (or a non-RegExp object) where the API contract promises a loud `SyntaxError`. That's the opposite of what the validation intends, and it fails silently rather than loudly.

### Suggested fix

Use `Object.create(null)` for the inner cache objects as well — two one-line changes:

```js
if (!patternCache[pattern]) {
    patternCache[pattern] = Object.create(null);
}
```

```js
XRegExp.cache = (pattern, flags) => {
    if (!regexCache[pattern]) {
        regexCache[pattern] = Object.create(null);
    }
    // ...
};
```

(Alternatively — or additionally — running `prepareFlags` before the cache lookup would make validation independent of cache state.)

### Suggested regression tests

```js
['__proto__', 'constructor'].forEach((flag) => {
    expect(() => XRegExp('.', flag)).toThrowSyntaxError();
    expect(() => XRegExp.cache('.', flag)).toThrowSyntaxError();
});
```

The existing suite covers reserved names in the *capture group name* position (`length` / `__proto__` in `tests/spec/s-xregexp.js`), but there is currently no coverage for reserved strings in the *flags* position, which is how this slipped through the #362 fix.

Happy to send a PR with the fix + tests if that's helpful. Thanks for the library!

simonyang08:
Fixes #365

## Summary

`new XRegExp(pattern, flags)` (and `XRegExp.cache(pattern, flags)`) stored their per-pattern sub-caches as plain objects (`patternCache[pattern] = {}`, `regexCache[pattern] = {}`). When `flags` is a name that exists on `Object.prototype` — `__proto__`, `constructor`, `hasOwnProperty`, `toString`, `valueOf`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString` — the lookup `patternCache[pattern][flags]` resolves through the prototype chain and is truthy, so the `Unknown regex flag` validation in `prepareFlags` is never reached. The constructor then falls through to `new RegExp(undefined, undefined)`, which produces the always-matching `/(?:)/`.

This is the same prototype-pollution class as the pattern caches addressed in #362, but for the second-level cache slots, the `fixed` table, and the `registeredFlags` lookup table.

## Changes

- `src/xregexp.js`: switch the affected caches and tables to `Object.create(null)` so prototype keys cannot silently satisfy the cache-miss check:
  - `const fixed = {}` → `Object.create(null)`
  - `const registeredFlags = { … }` → `Object.create(null)` + per-key assignment
  - `patternCache[pattern] = {}` → `Object.create(null)`
  - `regexCache[pattern] = {}` → `Object.create(null)`
- `tests/spec/s-xregexp.js`: add a Jasmine regression spec that drives all Object.prototype method names (plus `toLocaleString`) through `XRegExp('', name)` and asserts it throws a `SyntaxError`. On `master` these calls returned `/(?:)/` without throwing.

## Verification

- `npm test`: 420 specs, 0 failures (419 passing on master; +1 spec that fails there).
- `npm run lint`: clean.

## Notes

- No API changes for valid inputs; previously accepted inputs that produced `/(?:)/` now correctly throw a `SyntaxError`.
- The shape of the fix mirrors #362 (838ce76) exactly; only the affected storage and lookup sites differ.
- Follow-up noticed while auditing: `XRegExp.isInstalled` reads a plain-object `features` table, so `isInstalled('toString')` etc. wrongly returns true. Internal reads all use fixed keys, so it is unrelated to #365 — happy to file a separate fix.

Signed-off-by: simonyang08 <ppt5928@gmail.com>


slevithan:
Your PR is still marked as a draft so I didn't merge it, but I've accepted your approach in a separate diff.

